Issue #150, opened 2014-12-19
Multiple "instances" per module


teto:
Let me detail what I mean by "instance": for example in the clock module, by default with the mousewheel we can change the format.
I wish it were possible to do the same with :
- the mail module (ie switch between accounts with a click or mousewheel, easy with https://github.com/enkore/i3pystatus/pull/149) 
- or the network module (switch between interfaces to see their respective IPs).
- or media players. For instance I use mpd for music and vlc for audio. I wish i could switch between mediaplayers (i3pystatus could even detect which one is playing and display controls according to focus) while having both of them at the same place on the bar.

I wish it is clear and wonder if it interests others or if it looks too crazy.

cprn:
Isn't it possible by adding the `on_click` callback that would unregister one module and register the other? I didn't check but it seems doable.

PR #153, opened 2015-01-03
Cycle available interfaces on click

facetoe:
Cycle through available interfaces on click as suggested in #150.

### DIFF
--- a/i3pystatus/network.py
+++ b/i3pystatus/network.py
@@ -1,6 +1,4 @@
 from itertools import zip_longest
-import subprocess
-
 import netifaces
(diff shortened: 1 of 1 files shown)


teto:
when having lots of interfaces (as I sometimes have), it would be nice to have a next_interface / previous_interface rather than one function. Love the feature though :)

facetoe:
@teto That's a great idea. I'll have a go at implementing it this weekend.

teto:
You can see how I did in clock.py; callbacks accept extra parameters.

PR #154, opened 2015-01-03
Centralize network related functionality under one module.

facetoe:
I think it makes sense to group all the network related information under one module. It keeps all the network code together, provides a single place for configuration and allows the cycle on click feature to display all the new interface's information when cycling through. 

Merging this commit mean that users need to update their configs as it removes network_graph.py, network_traffic.py and wireless.py. The settings are also slightly different with the `hide_down` setting from NetworkTraffic being dropped and `format` being replaced by `format_up` and `format_down`. Apart from those changes I think everything should work as it did before. 

If you aren't opposed to merging the network stuff into one module I might do the same for the cpu ones when I get the chance.


enkore:
I like the idea, the separation of network, wireless and so on annoyed me quite a bit and wasn't really reasonable.

Some minor nit-picking:
- I don't like `NetUtil` with it's bunch of `@staticmethod`s. There's imho nothing wrong with having these methods on the module level
- I saw some classes deriving from `object`, which isn't necessary in Python 3 anymore
- Deriving from `ColorRangeModule` implies the colour dependency ; since `ColorRangeModule` is in core we just might bundle colour with i3pystatus to solve this.

teto:
I don't like NetUtil with it's bunch of @staticmethods either.
As a general remark, the problem does not lie with i3pystatus but rather the python community that does not provide any trustable network library. My favorite approach would be the libnl python bindings that returns results from the kernel via netlink, without error prone parsing. But the bindings are bloated (I've had patches for them but no time to go through with the upstreaming). As this PR makes the transition to another library easier, I support it.

facetoe:
Not sure what I was thinking when I added that class but I agree it serves no purpose. I've removed it along with the explicit inheritence from object.

teto:
I've upgraded to the master branch and I had to install 3 additionnal packages "psutil", "basiciw" and "basiciw-dev" for the same functionalities as before (display local wired ip):
```status.register("network", format_up="{v4cidr}",)```. Surely it's possible to display the error message only if the package is really needed ?

facetoe:
I'll have a look. I'm sure I can include the packages in a tighter scope so an error will only be thrown if you try and use that functionalty.

PR #161, opened 2015-01-11
Don't require importing psutil and basiciw unless necessary. 

facetoe:
Assorted bits and peices: 

* Prevent importing modules unless the functionality they offer is being used as per the comment on #154.
* Allow users to scroll through interfaces as suggested in #153.
* Added a module for checking Github notifications. 
* webrowser.open() was crashing on my system due to printing to stdout. Workaround redirects stdout temporarily to prevent this. 
* Added method to Reddit module to open the inbox.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -30,6 +30,7 @@ MOCK_MODULES = [
     "i3pystatus.pulseaudio.pulse",
     "notmuch",
     "requests",
+    "beautifulsoup4"
 ]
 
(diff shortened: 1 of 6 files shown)


enkore:
Could you rebase upon 611415c?

facetoe:
No worries, I forgot that the whosonlocation module requires beautifulsoup4 so I tried to add that as well. Looks like it didn't work though. I have to go to work now but if you want I can fix the failed build when I get home.

PR #162, opened 2015-01-15
Add action for open thunar

ismaelpuerto:


enkore:
With #149 you can do this directly in configuration. (Yes, we need to work on documentation regarding this topic)

Issue #165, opened 2015-01-21
[RFC] Provide option to use keyring to store sensitive credentials.


facetoe:
Several modules require username/password pairs, such as the reddit, github and mail modules. Storing these credentials in the clear is bad security. It would be nice if we could offer the option to use [keyring](https://pypi.python.org/pypi/keyring) instead to make things a little more secure. 

It is already possible to do this in the config, however some users may not know of the keyring module and integrating its functionality into i3pystatus would make it easier for them to take advantage of it. 

Implementing this would be easy. A setting could be added to each of the modules that require credentials to indicate that the keyring should be used instead of user/pass pairs. The only issue then is setting the credentials for the first time. If we agreed on a format such as `modulename.credential` we could provide instructions for setting them via the Python shell, or provide a small utility to help users set the values.

Let me know what you guys think.

teto:
I sync my dotfiles via git and as such is concerned by your point. I didn't know keyring but it looks like a nice solution. I personally appreciate programs that look into "$HOME/.netrc" to retrieve their password (like offlineimap and others). There does not seem to be a corresponding backup for keyring :'(

facetoe:
It looks like it is possible to write custom keyring backends and to specify at runtime which one to use. It should be pretty straightforward to write one that looks in "$HOME/.somedir" to retrieve credentials if you wanted to have copies of your credentials somewhere.

enkore:
One could handle this by making the `password` options optional and using keyring to retrieve the password if no `password` is specified. Or we could use a special sentinel object (i.e. at some prominent module `KEYRING=object()`) that can be specified as a password.

facetoe:
I like the idea of integrating this into the settings system. I'll play around a bit when I get the chance and try to get a prototype running for you guys to check out.

teto:
If it can inspire you https://github.com/untitaker/vdirsyncer/issues/14

facetoe:
Nice, that looks good. Can't hurt to support both keyring and netrc. Good find!

PR #166, opened 2015-01-25
Credentials

facetoe:
Added possible implementation of securing credentials for #165 and a hacky script to set credentials for modules that require them.

This implementation supports custom backends that the user can write or we could package with i3pystatus. All modules that would benefit from secure credentials will need a new class variable `keyring_backend` in order to support custom backends. If this value is defined then it will be used to retrieve credentials, otherwise the default backend will be used. Users can define these backends in their config, eg:

        status.register('github',
                        keyring_backend=NetrcBackend())

The key used to identify credentials is in the form `package.module.Class.setting`, eg `i3pystatus.reddit.Reddit.username`, or `i3pystatus.reddit.Reddit.password`. Any custom backends need to conform to this format in order for their credentials to be found. 

A setting is considered protected and will be searched for if it is in the `__PROTECTED_SETTINGS` array located in `SettingsBase` and is not already set. 

Let me know what you guys think. If you're happy with this and don't have any suggestions I'll update the docs and clean things up.


teto:
I will do my best to review it this week. Meanwhile do you think it would be possible to upstream the NetRc backend to keyring ? It would be in the interest of everyone, there is an issue open here https://bitbucket.org/kang/python-keyring-lib/issue/139/support-netrc-files .

teto:
I've added ```from i3pystatus.core.netrc_backend import NetrcBackend``` at the top of myStatus.py then I added:
```status.register('github', username="teto", keyring_backend=NetrcBackend()) ```
but it fails with invalid option 'keyring_backend'

**Comment by facetoe** [comment]
@teto I haven't yet added the `keyring_backend` option to all the modules, if you add the setting to the `Github` module it should work fine. Also I realized there is a bug where username/password entries in the `required` tuple aren't picked up so I need to fix that...

I'l have a look at cleaning up the NetrcBackend. I hacked that version together pretty quickly so it needs some work still.

**Comment by facetoe** [comment]
@enkore Any comments on this? Let me know if you want me to change something.

**Comment by enkore** [comment]
Sorry, I didn't notice that you added commits after your last comment. If you think it's good I'd merge it. Although some docs would be nice :)

**Comment by teto** [comment]
I've pulled your changes and tried the gihub module again with the Netrc backend but it still fails saying "Module 'Github': invalid option 'keyring_backend' " . Indeed it's not part of settings:
```
    settings = (
        ('format', 'format string'),
        ('unread_marker', 'sets the string that the "unread" formatter shows when there are pending notifications'),
        ("username", ""),
        ("password", ""),
        ("color", "")
    )
```
also I noticed a problem when losing internet connection, the github module generates a very long error that overlaps over the workspace buttons and then you can't  click on them. I believe it's more a problem with i3bar though (I opened a ticket on i3 for that).

enkore:
> the github module generates a very long error that overlaps over the workspace buttons and then you can't click on them.

Yeah, that's a long-standing bug with i3bar. In my defense, i3bar already had bugs with long status lines before I added click events.

facetoe:
Ahh, my bad. I added the variable but didn't add it to the settings tuple. Tomorrow I'll clean things up, write some docs and hopefully get these changes merged. Also that NetrcBackend is more a proof of concept than a solid implementation. I don't use .netrc files so it's not tested very well.

With the github module, can you remember the content of the error at all? Right now when the API returns an error a ConfigError is raised with the message returned by the API. It could be that the message is very verbose in which case I could truncate it so it doesn't take up the whole bar.

teto:
Truncating the output of errors could be a nice addition while waiting for some fix in i3bar (it may go in a separate PR).
I've got a 'Bad credentials' though my credentials look good. If you don't use it there is no reason for you to work further on this. You could remove it from the PR, and I will try to upstream the NetRcBackend in python-keyring but that won't be before the end of this month or next one.

facetoe:
Ok I've removed the NetrcBackend and added some documentation. Please let me know if you think the documentation is sufficient, and if not, where it could use more detail.

teto:
Finally I got some time to look through the code but it looks really complex, apparently you try to retrieve username/service/password form keyring while we only need to retrieve the password. Also if we set a keyring it's not used since it is checked before the module sets the content in 'settings' at members of the class. And to stick with i3pystatus conventions, it should be set as a string : "keyring.backends.gnome" instead of an instantiated class that forces to import classes from 'myStatus.py' (for instance). I've tried using a netrc backend but the service name it expects (is a domain name) looks different from what other backends expect. Which backend do you use and what does it expect as a 'service' ?

facetoe:
I don't think it's that complex. The basic idea is to check if a protected setting is already set, if not, try and set it. I'm not sure what you mean about retrieving username/service/password, I believe it is necessary to specify system/user/password when setting a credential, however you only need system/user to retrieve one. In this case "system" would be a string in the form `package.module.Class.setting`, however that is arbritrary and open to discussion. It might be worth checking out the  documentation on keyrings project page for more information on keyring itself - https://pypi.python.org/pypi/keyring. 

I'm just using the default keyring at the moment:

		>>> import keyring
		>>> keyring.get_keyring()
		<keyring.backends.file.PlaintextKeyring object at 0x7ff5845b4160>

However it shouldn't matter which one is used as they all have the same interface. I'm happy to look at defining the keyring as a string and then importing it, although I remember I ran into issues with this before when trying to use multiple different keyrings and setting it as an instantiated class seemed the simplest and most flexible option. Saying that though I'm not sure if anyone would actually want to use multiple keyrings so it could be a non-issue.  

I suspect a lot of the trouble you've been having stems from trying to use the `NetrcBackend`. As I mentioned before it's not at all well tested and the format of those files doesn't fit well with keyring. You would need to modify your files in order to conform to the format which i3pystatus expects, which is `package.module.Class.setting`, or better yet, modify the `NetrcBackend` to do a better job of things. 

Thanks for pointing out the bug where custom backends will never be used, I refactored things to be able to set credentials in the `required` tuple and introduced that bug in the process. I'll have a look tomorrow night and see if I can fix it.

facetoe:
@teto I've fixed the bug that prevented user-defined keyrings being used. I also refactored a little bit so hopefully it is more clear what is going on.

teto:
I will check that this evening. I've looked into gnome-keyring yesterday (to compare with netrc backend) and what I understood it that works the same, just that service names are different (I could not set the service associated with a password from "seahorse", I had to do it from python, which means it is the application that sets the service when registering a password ?!). 
My understanding is that depending on the backend, the service name changes (no format for gnome-keyring, domain name for netrc,  "package.module.Class.setting" for plainfiletext). So the service_identifer should be configurable to support other backends than PlainText.
If it is the case, then it brings little benefit to have it directly in i3pystatus since it becomes as complicated for the user to configure as to write the code to retrieve the password from "hisStatus.py". Any thought @enkore ?

facetoe:
@teto AFIK the service string is arbritrary, it's just an identifier that needs to be unique. The `package.module.Class.setting` that I chose to use for i3pystatus should work with any backend, the only exception being the prototype `NetrcBackend` that I wrote. The reason it doesn't work there is becuase .netrc is a totally different mechanism for storing passwords that just doesn't work well with keyring. From my understanding .netrc files are for storing credentials for FTP servers and the like while keyring is a general purpose key/value store. 

For most users configuration is dead easy. They run setting_util.py and set the credentials for their chosen module, that is all they have to do. If they are using a custom keyring (which would be rare), they simply need to set the service string in the format mentioned above and set the `keyring_backend` variable to their chosen module in the config. None of this seems very hard to me.

facetoe:
@enkore I think this is ready to be merged if @teto doesn't have any objections.

PR #167, opened 2015-01-27
Braille graphs

squidbeam:
I added options to cpu_usage_graph and network to draw braille graphs.

I did this by modifying make_graph to take a "style" parameter ("blocks" for yours, "braille" for braille), and then adding the above two corresponding options.

Your blocks have greater vertical resolution, but the braille has greater horizontal resolution.
braille

To save you some time considering this, I attached an image (with duplicated cpu plots for comparison), and here are the configuration lines I used:
            graph_style = 'braille',
            graph_width = 32,

### DIFF
--- a/i3pystatus/core/util.py
+++ b/i3pystatus/core/util.py
@@ -365,18 +365,23 @@ def internet():
         return False
 
 
-def make_graph(values, upper_limit=100.0, style="blocks"):
+def make_graph(values, lower_limit=0.0, upper_limit=100.0, style="blocks"):
     """
     Draws a graph made of unicode characters.
 
     :param values: An array of values to graph.
-    :param upper_limit: Maximum value for the y axis.
+    :param lower_limit: Minimum value for the y axis (or None for dynamic).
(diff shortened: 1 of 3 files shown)


facetoe:
Nice! Do you think it's possible to make a line graph with these characters?

squidbeam:
Yes, no problem.

I added two new choices, one that just draws the peak, and one that tries to fill things in a little.  Please see attached image.
braille2

facetoe:
Awesome! Yes if you want to add a lower limit to `make_graph` go right ahead. There is no harm in it if you set the default value to something sane.

squidbeam:
Great, I've added a commit with lower_limit.

There was an obvious default value of 0.0, which was already implicitly
used in cpu/network graphs when choosing the color.  I have additionally
added "None" as limit option to do the rescaling as before.  I did not
expose lower_limit in the cpu/network graph code since the default felt
fine and it seemed a bit much.

Thanks for suggesting line drawing: I personally like the
'braille-snake' graphs the most.

Issue #173, opened 2015-03-01
cpu_usage module ignores interval specification


stfuchs:
In `cpu_usage.py` the `init(self)` method seems to overwrite any interval specification I make during registration. Is this intended behaviour?

enkore:
Probably an oversight, non-standard intervals are usually specified in the class body.

Arvedui:
ooops my bad sorry :/
will submit a patch today

PR #174, opened 2015-03-01
remove the  hard coded interval in cpu_usage

Arvedui:
fix for #173


enkore:
Might want to change default interval to 1 still, in the class body

PR #175, opened 2015-03-01
minor improvment

tyjak:
Some minor improvments : 
* fix infinite percentage (backlight.py & network.py)
* add an option to choose your own Bitcoin symbol (bitcoin.py)


i3pystatus/network.py
```diff
@@ -156,6 +156,7 @@ def extract_wireless_info(self, interface):
             info["quality"] = quality["quality"]
         info["quality"] *= 100
         info["quality_bar"] = make_bar(info["quality"])
+        info["quality"] = '{:.0f}'.format(info["quality"])
```
enkore: This breaks existing configurations with float formatting. better use `round` here.

PR #183, opened 2015-03-12
Try/Except the pywapi call

djupp:
Often, the pywapi.get_weather... call will yield a "'NoneType' object has no attribute 'data' error" if the weather call does not work properly. 
This is usually resolved on the next call, but leads to an ugly warning in the status bar.
I simply wrapped the pywapi call in a try-except block where the except condition returns False. (Perhaps instead it should output something to a log file?)
This solved the problem for me.

PS: This is my first attempt at proposing a change, so if I'm not doing it right I'd love to be pointed to an intro to contributing to Github projects or something similar.

### DIFF
--- a/i3pystatus/weather.py
+++ b/i3pystatus/weather.py
@@ -42,7 +42,10 @@ class Weather(IntervalModule):
 
     @require(internet)
     def run(self):
-        result = pywapi.get_weather_from_weather_com(self.location_code, self.units)
+        try:
+            result = pywapi.get_weather_from_weather_com(self.location_code, self.units)
+        except AttributeError:
+            return False
         conditions = result["current_conditions"]
         temperature = conditions["temperature"]
         humidity = conditions["humidity"]


teto:
You can log with self.logger.log() (inherited from SettingsBase). I am not sure hiding by default the failure is something positive. That could be configurable via a setting (color_error/color_warning/hide_failure ...)

facetoe:
I've had similar problems with other modules that talk to the Internet. The `@require(internet)` decorator checks that there is Internet when beginning the call, but if you're like me and have unreliable Internet it can drop out during the `run()` method and raise an exception. 

I agree with @teto that just hiding the failure might not be the best solution, but maybe we could have a failure threshold of some sort where if the module failes more than N times an error is shown otherwise the error is suppressed?

djupp:
That sounds smart. 
Any chance you could hint at the canonical way to store variables across many runs of the same plugin? I have a bit of time over the next week to look at it, and I'll update the request accordingly once I've got it working.
By the way, I feel like with weather.com specifically it couldn't possibly be my connection that leads to the error messages (Ethernet at a research university - i.e. almost no dropped packages).

facetoe:
From memory the interval modules are instantiated once on load and then the `run()` method is called every `interval` seconds, so you should be able to set a variable in the module to keep track of the failures.

simon04:
The `AttributeError: 'NoneType' object has no attribute 'data'` error has been reported to `pywapi` and a patch is pending: https://code.google.com/p/python-weather-api/issues/detail?id=43

PR #187, opened 2015-03-18
Make mouse actions on the mpd module configurable

analogbyte:
I found the default mapping of mouse wheel up/down unintuitive, as the current default makes scrolling down go up in the playlist and the other way round. This is just a quick fix to make those values configurable. Thanks!


teto:
IIRC the on_*click options are inherited so you don't need to declare them in subclasses, ie. your additions are functionnally useless. You are correct though that it lacks documentation (like what are the available callbacks) and your PR fixes just that. I may give it a shot when I clean my PR to handle exotic buttons (such as mouse button 7/8)

analogbyte:
When I tested it it did complain about invalid options, and this addition makes that work. But yes, pulling those on_* options through on a general basis would be the best.

enkore:
Fixed since e41d1207

PR #190, opened 2015-03-28
Truncate long error messages.

facetoe:
Some module have very verbose error messages which fill up most of the screen. These commits allow adding a `max_error_len` variable to a module to truncate error messages longer than this value.


facetoe:
Not sure why the Travis build fails on the colour module, it looks like it installs it from pip...

PR #203, opened 2015-05-06
Add support for multiple batteries

mikn:
Some modifications I made locally to add support for multiple batteries (as present in the Thinkpad T440s and T450s), if you use the new default battery_ident "ALL" it will summarise all batteries. You can still set the battery manually by adding battery_ident to the keyword argument list when registering the status.
It also adds a new format output `no_of_batteries` which displays the count of batteries currently present.
It supports hotswapping of batteries. :)

### DIFF
--- a/i3pystatus/battery.py
+++ b/i3pystatus/battery.py
@@ -151,7 +151,7 @@ class BatteryChecker(IntervalModule):
         ("charging_color", "The charging color"),
         ("critical_color", "The critical color"),
         ("not_present_color", "The not present color."),
-        ("no_present_text", "The text to display when the battery is not present. Provides {battery_ident} as formatting option"),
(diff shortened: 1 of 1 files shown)


mikn:
Noticed I broke the tests quite badly, I'll be patching that soon-ish!

mikn:
So, the affected configuration options are as follows:
`path` Kept for legacy reasons to not break people's current configurations works the same way.
`not_present_text` Added a help text and that it now supports {battery_ident} as a formatting option.
New:
`paths` The paths for multiple batteries.
`base_path` The path where the batteries reside (defaults to `/sys/class/power_supply/`)
`battery_prefix` The prefix it should match against in the base_path to find batteries. It will look at the `uevent` in the root of this folder.

PR #204, opened 2015-05-07
Add scroll support for backlight

pandada8:


teto:
instead of using subprocess.Popen which in case of failure can throw an uncatched exception, I would recommand to use that https://github.com/enkore/i3pystatus/blob/master/i3pystatus/core/command.py.
If it gets merged, I will use this.

pandada8:
@teto thanks for you advice  :smile:

PR #205, opened 2015-05-20
Add module for monitoring OpenVPN connections.

facetoe:
Simple module to monitor OpenVPN status.

### DIFF
--- a/i3pystatus/openvpn.py
+++ b/i3pystatus/openvpn.py
@@ -1,0 +1,58 @@
+from i3pystatus import IntervalModule
+from i3pystatus.core.command import run_through_shell
+
+__author__ = 'facetoe'
+
+
+class OpenVPN(IntervalModule):
+    """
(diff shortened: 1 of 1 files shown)


facetoe:
Any idea why this build fails? I can't see the start string without an end string...

i3pystatus/openvpn.py
```diff
@@ -0,0 +1,58 @@
...
+    Formatters:
+
+    * {vpn_name} — Same as setting.
+    * {status} — Unicode up or down symbol.
+    * {output} — Output of status_command.
+    * {label} — Label for this connection, if defined.
+
+    """
+
+    colour_up = "#00ff00 "
```
enkore: `"#00ff00 "` <- it's the space at the end. Really obscure one :)

facetoe: Ahhh good spot!

PR #208, opened 2015-06-03
Initial commit of openstack_vms.py

drwahl:
This patch provides an module for tracking the number of active and
non-active VMs in a given openstack cluster.  When non-"ACTIVE" VMs
are above a given threshold, the text will (by default) change to
red, thus indicating an issue.  Otherwise, the text will be green.

### DIFF
--- a/i3pystatus/openstack_vms.py
+++ b/i3pystatus/openstack_vms.py
@@ -17,12 +17,10 @@ class Openstack_vms(IntervalModule):
         ("username", "Username for OpenStack authentication (OS_USERNAME)"),
         ("password", "Password for Openstack authentication (OS_PASSWORD)"),
         ("tenant_name", "Tenant/Project name to view (OS_TENANT_NAME)"),
-        ("color", "Display color when non-active VMs are =< `threshold` "
-            "(default: #00FF00"),
-        ("crit_color", "Display color when non-active VMs are => `threshold` "
-            "(default: #FF0000"),
+        ("color", "Display color when non-active VMs are =< `threshold`"),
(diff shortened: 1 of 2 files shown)


i3pystatus/openstack_vms.py
```diff
@@ -0,0 +1,69 @@
+from i3pystatus import IntervalModule
+from .core.util import round_dict
+# requires python-novaclient
```
enkore: Please put that in the module description below, too.

You also need to add `novaclient.v2` to the mocks list at https://github.com/enkore/i3pystatus/blob/master/docs/conf.py#L23 (yeah, this is a bit ... uhm... but it's necessary to allow automatic generation of the docs without actually *needing* all the modules installed.

i3pystatus/openstack_vms.py
```diff
@@ -0,0 +1,69 @@
...
+    non-ACTIVE states.
+    """
+
+    settings = (
+        ("auth_url", "OpenStack cluster authentication URL (OS_AUTH_URL)"),
+        ("username", "Username for OpenStack authentication (OS_USERNAME)"),
+        ("password", "Password for Openstack authentication (OS_PASSWORD)"),
+        ("tenant_name", "Tenant/Project name to view (OS_TENANT_NAME)"),
+        ("color", "Display color when non-active VMs are =< `threshold` "
+        "(default: #00FF00"),
```
enkore: You don't need to manually document the defaults, they are automatically documented.

drwahl:
Ok, I think I've finally gotten all the issues addressed.  Travis seems happy too.  Let me know if you find anything else amiss.

PR #209, opened 2015-06-03
Updates module

richese:
A generic update checker.
This is basically just stripped down mail module with different backends.

Current backends are:
* Pacman - Arch. Is using the checkupdates script.
* Cower - Arch. For AUR packages.
* AptGet - Debian. It's not ideal but it seems to work as intended. I only tested this on fresh install of Debian 8.

### DIFF
--- a/i3pystatus/updates/__init__.py
+++ b/i3pystatus/updates/__init__.py
@@ -19,6 +19,22 @@ class Updates(IntervalModule):
 
     * `{count}` — Sum of all available updates from all backends.
 
+    .. rubric:: Usage example
+
+    ::
+
+        from i3pystatus import Status
+        from i3pystatus.updates import pacman, cower
+
+        status = Status(standalone=True)
+
+        status.register("updates",
+                        format = "Updates: {count}",
+                        format_no_updates = "No updates",
+                        backends = [pacman.Pacman(), cower.Cower()])
+
+        status.run()
+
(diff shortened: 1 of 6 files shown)


facetoe:
This is awesome. I've tested the `Pacman` and `Cower` backends and they both work perfectly on my machine.

One thing that would be nice though is to be able to see how many updates are pending for each backend. Maybe a `name` variable could be added to the backend and an option could be provided to show the label before each count. Or maybe you could scoll through total, and then each backend. 

Anyway nice work!

richese:
I would rather solve that by using multiple modules in config file to keep the code simple. Usage of multiple backends is Arch Linux specific thing, since repos and AUR are separated.

i3pystatus/updates/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+
+        status.run()
+
+    """
+
+    interval = 3600
+
+    settings = (
+        ("backends", "Required list of backends used to check for updates."),
+        ("format", "String shown when updates are availible. "
```
enkore: "available" instead of "availible"

enkore:
> This is awesome. 

Indeed very nice work!

facetoe:
@richese Ahhh good point about the multiple backends being Arch specific. Mutliple instances it is!

PR #212, opened 2015-06-04
makewatch: a module to watch for make (or other long running) jobs

drwahl:
A module that will watch for (by default) make jobs and notify of their
status.  This can be used for other long-running processes by providing
an alternate 'name'.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -31,8 +31,7 @@ MOCK_MODULES = [
     "notmuch",
     "requests",
     "bs4",
-    "novaclient.v2",
-    "psutil", "getpass"
+    "novaclient.v2"
 ]
 
 for mod_name in MOCK_MODULES:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,7 +32,7 @@ MOCK_MODULES = [
     "requests",
     "bs4",
     "novaclient.v2",
-    "dota2py",
+    "psutil", "getpass"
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -31,7 +31,8 @@
     "notmuch",
     "requests",
     "bs4",
-    "novaclient.v2"
+    "novaclient.v2",
+    "psutil", "getpass"
```
enkore: - getpass is part of the Python standard library
- psutil is already on the list (line 25)

i3pystatus/makewatch.py
```diff
@@ -0,0 +1,44 @@
...
+
+        cdict = {
+            "name": self.name,
+            "status": status
+        }
+
+        self.output = {
+            "full_text": self.format.format(**cdict),
+            "color": color
+        }
```
enkore: An option to hide it when idle would be nice

drwahl:
I've been playing with getting the format to change dynamically (including "hiding"), but have not had success yet.  Are you able to provide me some pointers here?

facetoe:
You can use the `on_leftclick`, `on_rightclick` callbacks to trigger actions in your module. Hiding could be as simple as changing the output to an empty string.

enkore:
Cherry-picked relevant commits into 7be1361 8b8a3db b09e213

PR #213, opened 2015-06-05
CPU frequency module

gacekjk:
It shows cpu frequency based on /proc/cpuinfo.
Basic tests included.

### DIFF
--- a/i3pystatus/cpu_freq.py
+++ b/i3pystatus/cpu_freq.py
@@ -1,0 +1,25 @@
+from i3pystatus import IntervalModule
+
+
+class CpuFreq(IntervalModule):
+    format = "{avg}"
+    settings = (
+        "format",
+        ("color", "The text color"),
+        ("average", "Show average for every core"),
+        ("file", "override default path"),
+    )
+
+    file = '/proc/cpuinfo'
+    color = '#FFFFFF'
+
+    def run(self):
+        with open(self.file) as f:
+            mhz_values = [float(line.split(':')[1]) for line in f if line.startswith('cpu MHz')]
+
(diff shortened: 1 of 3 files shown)


i3pystatus/cpu_freq.py
```diff
@@ -0,0 +1,54 @@
+"""
+The module gathers information by default from `/proc/cpuinfo` about the current cpu frequency
+"""
+# coding=utf-8
```
enkore: Not necessary with Python 3

i3pystatus/cpu_freq.py
```diff
@@ -0,0 +1,54 @@
+"""
+The module gathers information by default from `/proc/cpuinfo` about the current cpu frequency
```
enkore: This docstring should go into the class docstring below ; the docstring of Python modules is ignored.

i3pystatus/cpu_freq.py
```diff
@@ -0,0 +1,54 @@
...
+class CpuFreq(IntervalModule):
+    """
+    class uses by default `/proc/cpuinfo` to determine the current cpu frequency
+
+    .. rubric:: Available formatters
+
+    * `{avg}` - mean from all cores in MHz `4.3f`
+    * `{avgg}` - mean from all cores in GHz `1.2f`
+    * `{corex}` - frequency of a selected core in MHz `4.3f`
+    * `{corexg}` - frequesncy of a selscted core in GHz `1.2f`
```
enkore: It's not very clear from the description that `x` is just a placeholder here, maybe rephrase it into `{coreX} - frequency of core X in ...` or so

enkore:
Just some minor things, looks good.

gacekjk:
Hopefully right now it's better :smile:

PR #215, opened 2015-06-08
net_speed: a module to display internet speeds

drwahl:
This module provides a glimpse/snapshot of current internet based
speeds.  To keep things simple (and due to bugs in speedtest_cli), this
module simply uses requests to download a small(ish) image and times the
download.  This is not a perfect/definitive test, but it should be a
decent indicator of how the internet bound traffic looks on the clients
network.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,7 +32,8 @@ MOCK_MODULES = [
     "requests",
     "bs4",
     "dota2py",
-    "novaclient.v2"
+    "novaclient.v2",
+    "speedtest_cli"
 ]
 
 for mod_name in MOCK_MODULES:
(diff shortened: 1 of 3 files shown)


enkore:
Cherry picked into 9c246bc6 3ad6fc49

Small bit of advice: try to keep the commits in a PR "clean". Having upstream commits in a PR is often caused if you `git-pull` upstream to update your fork, but didn't use `--rebase`, which is what you want in this case (to re-base your changes upon the updated upstream).
While there is no *technical* issue per the with having upstream commits in a merge (git handles it fine and correctly), it creates a rather confusing commit history with lots of seemingly duplicated commits getting merged "into each other".

PR #218, opened 2015-06-16
Dota2stats

drwahl:
In order to avoid filling i3pystatus with super long decimals, this
patch sets the dota2win module to use only 2 decimal places for it's win
percent.


enkore:
Cherry picked to b2175b19311a

PR #219, opened 2015-06-17
Add module for diplay keyboard layout

ismaelpuerto:


enkore:
A while back I recommended someone this way:

    status.register("shell", command="""setxkbmap -query | awk '{ if($1=="layout:") print $2 }'""", interval=10)

If we'd want this in `master` one could just do it with a subclass of `Shell`

Issue #181, opened 2015-03-09
Ampersands in i3 4.9


a-roy:
I just updated i3 to the latest version (4.9.1) and now have an issue with ampersands causing modules in i3pystatus not to be displayed. Using the text module, I found that using `&amp;` correctly displays an ampersand. However, I can't make this replacement in other modules which use format strings to generate the text (such as the MPD module).

richese:
i3bar parses output strings either as normal text or as pango markup. Pango seems to require ampersands in the `&amp;` form or some nasty things happen.

AFAIK there is currently no way to set output to pango markup mode in i3pystatus so I am not sure why this happend in the first place. Probably some changes on i3bar side.

I just found this randomly while testing some things that allow the use of pango. This bug does not show for me otherwise.

One possible solution could be to explicitly add `"markup": "none"` to output of each module or `"markup": "pango"` and replace the ampersands.

enkore:
Fixed by #220

PR #220, opened 2015-06-18
Added hints setting to Module.

richese:
This setting gives acces to all features of i3bar protocol that are not handled by most modules but may be useful sometimes.

Rigth now this PR will add `"markup": "none"` to all outputs by default. This should prevent #181 from  happening but technically it should be the default anyway so I am not sure if to keep it there.


enkore:
Looks good, but the terminology isn't right. The output of a module in i3pystatus is the same as an i3bar block, the key/value pairs of the block are not blocks in their own regard but just key/value pairs. It's a bit misleading in the i3bar docs that section 2.2 is titled "Blocks in detail", what they mean is the structure of blocks in detail, i.e. permitted keys and how the values of those keys should look like.

richese:
I see what you mean. I'll rewrite & move the docs.

i3pystatus/core/modules.py
```diff
@@ -13,6 +13,7 @@ class Module(SettingsBase):
         ('on_rightclick', "Callback called on right click (see :ref:`callbacks`)"),
         ('on_upscroll', "Callback called on scrolling up (see :ref:`callbacks`)"),
         ('on_downscroll', "Callback called on scrolling down (see :ref:`callbacks`)"),
+        ('hints', "Additional output blocks for module output (see :ref:`hints`)"),
```
richese: These refferences are not displaying correctly in the [documentation](http://docs.enkore.de/i3pystatus/i3pystatus.core.html#i3pystatus.core.modules.IntervalModule.settings).

enkore: Yeah sphinx doesn't parse normal strings in the code. It does show correctly in the module reference, though.

Issue #155, opened 2015-01-04
Refresh the bar


Phyks:
Hi,

I set a high interval between refresh for most of my plugins, as I do not need realtime updates often.

But sometimes, it could be nice to have a direct realtime feedback, when changing brightness through keyboard keys for instance. Then, a nice solution would be to handle a `kill -USR1` and refresh the whole bar on such signal.

What do you think ?

teto:
...or when a cronjob updates your mailbox, it would be nice if it could refresh the display.

facetoe:
I think this would be a great feature. It would also be cool if click events could be handled immediately without waiting for the module to be called again.

enkore:
Implemented in #223

PR #222, opened 2015-06-20
Bar refresh

richese:
This PR implements features from #155.
* The whole statusline is refreshed if SIGUSR1 is recieved.
* After every click event with valid callback, modules output is refreshed by calling ``run`` and updated statusline is sent to i3bar.

Standalone operation is required for this features.

SIGUSR2 is used internally for click events. Recieving it flushes current state to i3bar immediately.

Now I am just wondering if it'll actually work fine with anyone else's config files.

Edit: Maybe adding option to disable specific module from refreshing could be useful?

### DIFF
--- a/i3pystatus/core/io.py
+++ b/i3pystatus/core/io.py
@@ -52,7 +52,7 @@ class StandaloneIO(IOHandler):
     and the i3bar protocol header
     """
 
-    __is_click_event = False
+    refresh_modules = True
 
     n = -1
     proto = [
@@ -77,11 +77,15 @@ class StandaloneIO(IOHandler):
                 return
 
             if received_signal:
-                if StandaloneIO.__is_click_event:
-                    StandaloneIO.__is_click_event = False
-                else:
+                if StandaloneIO.refresh_modules:
+                    # refresh whole bar
                     for module in self.modules:
                         module.on_refresh()
+                else:
+                    # just send status line to i3bar immediately -> do nothing here
+
+                    # set next signal action back to default
+                    StandaloneIO.refresh_modules = True
 
             yield self.read_line()
 
@@ -91,8 +95,12 @@ class StandaloneIO(IOHandler):
         return self.proto[min(self.n, len(self.proto) - 1)]
 
     @classmethod
-    def register_click_event(cls):
-        cls.__is_click_event = True
+    def refresh_statusline(cls):
+        """
+        Changes behavior of the next SIGUSR1 signal to just flushing current
+        outputs of all modules and sends the signal.
+        """
+        cls.refresh_modules = False
         os.kill(os.getpid(), signal.SIGUSR1)
 
 
--- a/i3pystatus/core/io.py
+++ b/i3pystatus/core/io.py
@@ -65,17 +65,19 @@ class StandaloneIO(IOHandler):
 
     def read(self):
         while True:
+            info = None
             try:
                 info = signal.sigtimedwait([signal.SIGUSR1, signal.SIGUSR2],
                                            self.interval)
-                # refresh the whole bar
-                if info and info.si_signo == signal.SIGUSR1:
-                    for module in self.modules:
-                        module.run()
             except InterruptedError:
                 logging.getLogger("i3pystatus").exception("Interrupted system call:")
             except KeyboardInterrupt:
                 return
+
+            # refresh the whole bar
+            if info and info.si_signo == signal.SIGUSR1:
+                for module in self.modules:
+                    module.on_refresh()
 
             yield self.read_line()
 
--- a/i3pystatus/core/io.py
+++ b/i3pystatus/core/io.py
@@ -56,16 +56,22 @@ class StandaloneIO(IOHandler):
         {"version": 1, "click_events": True}, "[", "[]", ",[]",
     ]
 
-    def __init__(self, click_events, interval=1):
+    def __init__(self, click_events, modules, interval=1):
         super().__init__()
         self.interval = interval
         self.proto[0]['click_events'] = click_events
         self.proto[0] = json.dumps(self.proto[0])
+        self.modules = modules
 
     def read(self):
         while True:
             try:
-                signal.sigtimedwait([signal.SIGUSR1], self.interval)
+                info = signal.sigtimedwait([signal.SIGUSR1, signal.SIGUSR2],
+                                           self.interval)
+                # refresh the whole bar
+                if info and info.si_signo == signal.SIGUSR1:
+                    for module in self.modules:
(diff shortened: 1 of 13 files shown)


facetoe:
You're on fire @richese, this is great!

One thing I noticed with my config is sending the `SIGUSR1` signal causes my whole bar to freeze while the long running modules (eg ones that call out to the Internet) complete. This can be fixed by queuing up the jobs and executing them in background threads, eg:

                # refresh the whole bar
                run_thread = threading.Thread(0)
                if info and info.si_signo == signal.SIGUSR1:
                    for module in self.modules:
                        run_thread.append(module.run)
                run_thread.start()

richese:
@facetoe Yes, thats the only problem I noticed. Using a thread sounds interesting, I'll try that for sure.
I considered creating an `on_sigusr` callback which defaults to `run` for fast modules and to `None` for slow (or it might decide whether to add it to the thread or execute without delay).

:question: Also, is it even safe to call ``run()`` in the first place? It just werks right now but something could possibly happen.

facetoe:
@richese I like the idea of the `on_sigusr` callback. It adds a bit of flexibility for what happens when a signal is recieved and leaves it up to the module developers to decide what they would like to happen. That said, in most cases just calling the module again would be fine, so it might be worth setting an option on those modules that want to ignore the signal, as you suggest above, eliminating the need for another callback. Maybe @enkore can weigh in here.

With calling run by default, it *should* be safe, although it is possible that some of the `IntervalModule` subclasses could return incorrect results if they are depending on the timing. I believe the `Network` module relies on the interval being one second in order to do some calculations.

richese:
The callback could look (roughly) like this for slow modules like `Updates`:
``` python
on_sigusr = "refresh"
def refresh(self):
    def refresh_job(module):
        module.run()
        os.kill(os.getpid(), signal.SIGUSR2)
        
    refresh_thread = threading.Thread(target=refresh_job, args=(self))
    refresh_thread.start()
```

enkore:
I'm okay with another callback, but I don't like the name (`on_update` or something like that is better IMHO, since `on_sigusr` is a bit cryptic what it's about). I also don't think this should be a setting, since this is around to avoid internal issues, nothing you'd need to customize.

That being said we can just make it a normal method.

> :question: Also, is it even safe to call run() in the first place? It just werks right now but something could possibly happen.

We neither guarantee thread affinity nor exact timing. We only say "okay, in the long run the interval gets close/is what you specified". But from call to call it can be e.g. 0.9 s, 1.05 s, 1.1 s, 0.9 s, 1 s, 1.05 s and so on (on average 1 s). 

>  I believe the Network module relies on the interval being one second in order to do some calculations. 

Yes, the NetworkTraffic class implicitly assumes a one second interval with calls to NetworkTraffic.get_usage.

Arvedui:
Why using a different signal for the refresh on click stuff?
Since it's send internally it could easily be done with some kind of multiplexing with sigusr1. That would keep sigusr2 free for other stuff that cannot not be solved that way like a user initiated in place restart.

richese:
@Arvedui Guess I was just too lazy after not finding a place where it would 'fit' nicely.

i3pystatus/core/io.py
```diff
@@ -50,24 +52,37 @@ class StandaloneIO(IOHandler):
     and the i3bar protocol header
     """
 
+    __is_click_event = False
```
enkore: Identifiers starting with `__` shouldn't be normally used (per PEP8) ; use `_` instead.

i3pystatus/core/io.py
```diff
@@ -50,24 +52,37 @@ class StandaloneIO(IOHandler):
...
-                time.sleep(self.interval)
+                received_signal = signal.sigtimedwait([signal.SIGUSR1], self.interval)
+            except InterruptedError:
+                logging.getLogger("i3pystatus").exception("Interrupted system call:")
             except KeyboardInterrupt:
                 return
 
+            if received_signal:
+                if StandaloneIO.__is_click_event:
+                    StandaloneIO.__is_click_event = False
```
enkore: Uh, not sure if I get this right.

_i_c_e is set (at class construction) to True, if it's true && received_signal, it's set to false. I don't quite grasp this - why is it necessary?

enkore:
I reviewed the code changes and (as posted above) I'm not quite understanding it all, it would be great if you could explain the logic a bit more, maybe we can even simplify it :)

richese:
Yeah, I hope this can be simpified to some extent.
There should be a separate class for signal handling but I have not figured out what it should look like yet. StandaloneIO is the unfortunate guy that gets the job done right now.

I added one more commit where I did some renaming (to better names hopefully) and added some docstrings.

To explain it. SIGUSR1 signal is used for three things:
* Update all modules outside of their scheduled intervals by calling their `run` method .
 * To prevent freezing of the whole bar some slower modules are refreshed in separate threads. When its output is updated SIGUSR1 is send again to show output of this module. To differ between 'slow' and normal modules the 'on_refresh' method is used.
* Update status line immediately after a click event. This implies that output of the clicked module has been updated already.

Where points two and three are the same thing.
To know which action to do, there must(?) be one global/class variable that tells what will happen on next signal. This is the `refresh_module` class variable in StandaloneIO.
When you want to flush current output to i3bar you call the `refresh_bar` class method in StandaloneIO which sets the variable temporarily and sends the signal.

I hope that explains it a bit. :)

enkore:
Thanks for the explanation.

I think the click-events related part could be removed, since we can do half of it already (or I'm not getting it right yet ;), by using callbacks e.g.

    status.register("clock", interval=60, on_leftclick="run")

The other half would be a bar update after a click event (to have instant feedback if something was changed by the click). Now we do this by sending USR1, which updates *all* modules / or not all modules, depending on the global variable.

Now I see what you used USR2 for before ("*The actual suspension time may be less than that requested because any caught signal will terminate the sleep() following execution of that signal’s catching routine.*").

A `threading.Condition` in the StandaloneIO might be a nicer way around this (to output updated module contents, just `notify()` it to resume the I/O thread. Timeout will do it inside specified interval otherwise.)

richese:
@enkore I tried implementing it with the Condition as you mentioned. It has it's own quirks but the code looks much better now. I will probably make another PR from different branch later when I sort it out.

> Regarding long-taking modules that should be updated from a separate thread it could either be done via a method or by inferring this from the interval – low interval implies short duration (e.g. clock, network, CPU, backlight, ...), long interval implies long duration of the action (e.g. mail, net_speed, ...).

I like this idea. By looking at my config, an average interval from all modules might be a sufficiently good treshold value.

PR #223, opened 2015-06-22
Bar refresh (take 2)

richese:
Continued, based on discussion from #222.

Hopefully with more descriptive names.

### DIFF
--- a/i3pystatus/core/io.py
+++ b/i3pystatus/core/io.py
@@ -130,14 +130,13 @@ class StandaloneIO(IOHandler):
         method.
         The reasoning is that modules with larger intervals also usually take
         longer to refresh their output and that their output is not required in
-        `real time`.
+        'real time'.
         This also prevents possible lag when updating all modules in a row.
         """
 
         if signo != signal.SIGUSR1:
             return
 
-        threads = []
         for module in StandaloneIO.modules:
             if hasattr(module, "interval"):
                 if module.interval > StandaloneIO.treshold_interval:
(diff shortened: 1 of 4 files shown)


i3pystatus/core/io.py
```diff
@@ -75,6 +93,62 @@ def read_line(self):
...
+        The reasoning is that modules with larger intervals also usually take
+        longer to refresh their output and that their output is not required in
+        'real time'.
+        This also prevents possible lag when updating all modules in a row.
+        """
+
+        if signo != signal.SIGUSR1:
+            return
+
+        for module in StandaloneIO.modules:
```
enkore: ok, let's make r_s_h a @classmethod instead of @staticmethod to have it have it consistently

i3pystatus/core/io.py
```diff
@@ -75,6 +93,62 @@ def read_line(self):
...
+    @classmethod
+    def compute_treshold_interval(cls):
+        """
+        Current method is to compute average from all intervals.
+
+        Make sure to call after all modules are registered.
+        """
+
+        intervals = [m.interval for m in cls.modules if hasattr(m, "interval")]
+        cls.treshold_interval = round(sum(intervals) / len(intervals))
```
enkore: I think calls to this are "rare enough" that we don't need to additionally cache it with a class-level variable

i3pystatus/core/io.py
richese: Oh, I didn't know I can do that and make it work. 

In that case I can remove all the static and class methods and make it either:
1. Singleton
2. Or pass it to CommandEndpoints constructor (as that is the only other point that needs to call `async_refresh`)

enkore:
Very nice - merged

PR #224, opened 2015-07-06
AnyBar widget

narma:
Hi!

It's a port of https://github.com/tonsky/AnyBar.

### DIFF
--- a/i3pystatus/anybar.py
+++ b/i3pystatus/anybar.py
@@ -1,0 +1,66 @@
+import threading
+import socket
+
+from i3pystatus import IntervalModule
+
+
+class AnyBar(IntervalModule):
+    """
+    This module shows dot with given color in your panel.
+    What color means is up to you. When to change color is also up to you.
+    It's a port of https://github.com/tonsky/AnyBar to i3pystatus.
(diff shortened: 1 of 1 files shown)


narma:
Oh, it requires python 3.4 for asyncio. It's ok or we need to support 3.3 too?

enkore:
I'm not exactly sure how to handle this. I think it would be ok to say "this needs either Python 3.4+, or 3.3+ with asyncio installed" (asyncio => https://pypi.python.org/pypi/asyncio). In this case, to fix the CI/build issue, see http://docs.enkore.de/i3pystatus/module.html#handling-dependencies (i.e. "asyncio" goes into MOCK_MODULES)

i3pystatus/anybar.py
```diff
@@ -0,0 +1,70 @@
...
+            self.__class__.loop_thread.start()
+
+    def run(self):
+        self.output = {
+            "full_text": "●",
+            "color": self.colors.get(self.color, self.color)
+        }
+
+
+def anybar_remove(status, anybar):
```
enkore: What is this for?

narma:
I've updated PR with using regular socket/threading approach.

PR #235, opened 2015-08-10
added the alert_command option

plumps:
running a shell command before the battery is completely discharged (~1% left).
You just need to add 

alert_command="<path_to_program> [$@]" 

e.g. alert_command="/usr/bin/systemctl hibernate"

### DIFF
--- a/i3pystatus/battery.py
+++ b/i3pystatus/battery.py
@@ -140,7 +140,7 @@ class BatteryChecker(IntervalModule):
         ("not_present_text", "Text displayed if the battery is not present. No formatters are available"),
         ("alert", "Display a libnotify-notification on low battery"),
         ("critical_level_command", "Runs a shell command in the case of a critical power state"),
-        "critical_level_percentage"
+        "critical_level_percentage",
         "alert_percentage",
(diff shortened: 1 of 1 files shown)


Arvedui:
I don't think that this should be implemented on this level.

It is the job of you favourite energy management tool to do this and not of a high level tool like a status bar generator.

plumps:
Which is? 

Unfortunately you cannot write udev rules for most systems, as the hardware seldomly generates those uevents. Therefore something has to poll the value of /sys/class/power_supply/* on a regular base, which should be done in a sparse manner.  And as i3pystatus has to do that anyway, I thought it'd be a nice place to trigger other events as well.

Arvedui:
I thought it to be kind of default functionality of tools like tlp.
But obviously I was wrong …

Could you add an config option for the threshold value?
There may be conditions where 1% is not enough and I can imagine that some people are just not comfortable with walking that close to the edge.

plumps:
Sure :)

plumps:
Here's what has changed:
- alert_command became **critical_level_command**
- **critical_level_percentage** to adjust the level, when the command gets triggered. default_value = 1

PR #245, opened 2015-09-10
Added moon phase module

Elder-of-Ozone:


i3pystatus/moon.py
```diff
@@ -0,0 +1,123 @@
...
+                                Last Quarter:
+                                Waning Crescent:
+                         }
+        """
+
+        settings = (
+            "format",
+            ("status", "Current moon phase"),
+            ("illum", "Percentage that is illuminated"),
+            ("color", "Set color"),
```
enkore: I think you mean `phase_color` here?

i3pystatus/moon.py
```diff
@@ -0,0 +1,123 @@
...
+            "First Quarter": "#265ECC",
+            "Waxing Gibbous": "#392FBF",
+            "Full Moon": "#4C00B3",
+            "Waning Gibbous": "#871181",
+            "Last Quarter": "#C32250",
+            "Waning Crescent": "#FF341F",
+        }
+
+        def pos(now=None):
+            now = None
```
enkore: now = None makes the parameter unnecessary

i3pystatus/moon.py
```diff
@@ -0,0 +1,123 @@
...
+            days_in_second = 86400
+
+            if now is None:
+                now = datetime.datetime.now()
+
```
enkore: And the if here, too, since it is always None at this point

Elder-of-Ozone: Ah thanks. I fixed your suggestions.

Issue #244, opened 2015-09-10
format_down() for mpd


stagnation:
I would like to have a specific format when the mpd server isn't reachable, as it stands all I see is the error.

MPD: [Errno 113] No Route to host.

status.register("mpd",
    format="{artist} : {title}",
    host=("192.168.1.9"),
    status={
        "pause": "▷",
        "play": "▶",
        "stop": "◾",
    },)

works great when I have the correct local ip though. but would rather want a discrete "no mpd" message than a very red error.

gutodisse1:
Please, make backup before. I'm newbie with free time, I don't expect boring you with more erros.

stagnation:
looks good to me but some small nitpicking.
here's the much better error handling: 
http://aelv.se/spill/ul/py3stat_error.png
and working as usual otherwise
http://aelv.se/spill/ul/py3stat_online.png (with full i3 restarts between these)

while writing this I turned off my internet connection to see what would happen if connection was lost
and so on. After reinstating the connection to the server and playing a song I got the correct display of playing song, but in the color_error color. (don't have a default color value in the cfg).

after setting a good color value and toggling internet connection mpd display doesn't update as connection is lost to server. (just shows last known message / song until connection is re-established.

cloned your patch-4 branch and copied the p3status dir to where my status script is located in terms of installation

*seems like the color key is needed in conjunction with error_color, sometimes error color is used when it is working.
*doesn't update when connection to mpd server is lost.

gutodisse1:
About color erros I patch this. 
About when you turn off internet, I need learn more about sockets - I don't know much about sockets.
The program check if your connection is open and check when received data, I guess your problem is when you open connection and don't receive nothing more, the program show the last information received.
(Sorry my English. I'm student)

enkore:
Is #251 an adequate solution for everyone?

PR #249, opened 2015-09-11
Update mpd.py

gutodisse1:
My first experiment. I try solve  issue #244. 
Can stagnation(User from issue) test and report me anything?


facetoe:
Looks like this one was resolved in #251.

PR #253, opened 2015-09-16
Fixed a race condition which sometimes caused an empty output.

schroeji:
There was a race condition in the cmus module which caused an empty output string from time to time.
I fixed it.


t3zla:
Please make it pass the Travis Cl build

Issue #254, opened 2015-09-21
i3pystatus blocks after application has started


joepvd:
If a mouse action is set to run a long running application, clicks on the status bar are ignored until the application is closed. This can be observed wit hthe default settings when the pulseaudio plugin is enabled. From what I casn tell, after opening the a pavucontrol application from i3pystatus, subsequent clicks on this and other modules are being acted upon. As soon as the blocking application is closed, the clicks are replayed. 

**Expected behavior**
Processes are spawned fork/exec-ed from the i3pystatus process, and i3pystatus does not care if this process runs, if it exited, and if so, erroneously or not.

enkore:
Fixed in merged PR #257

PR #257, opened 2015-09-23
Fix blocking clickevents

richese:
Added new function `execute` to run external programs without waiting for their output (#254).

Also some additions to `run_through_shell`:
* Arguments are separated automatically if shell is not enabled. This way you can simply pass the string with the command without using `command.split()` every time you want to run something without using shell.
* Exceptions are now logged. Just in case they ever actually happen.

### DIFF
--- a/i3pystatus/core/command.py
+++ b/i3pystatus/core/command.py
@@ -25,13 +25,14 @@ def run_through_shell(command, enable_shell=False):
      string since shell does all the parsing.
     """
 
-    if not enable_shell and not isinstance(command, list):
+    if not enable_shell and isinstance(command, str):
         command = shlex.split(command)
 
     returncode = None
     stderr = None
     try:
-        proc = subprocess.Popen(command, stderr=subprocess.PIPE, stdout=subprocess.PIPE, shell=enable_shell)
+        proc = subprocess.Popen(command, stderr=subprocess.PIPE,
+                                stdout=subprocess.PIPE, shell=enable_shell)
         out, stderr = proc.communicate()
         out = out.decode("UTF-8")
         stderr = stderr.decode("UTF-8")
@@ -64,10 +65,13 @@ def execute(command, detach=False):
 
     if detach:
         if not isinstance(command, str):
-            raise TypeError("Detached mode expects a string as command.")
+            msg = "Detached mode expects a string as command, not {}".format(
+                  command)
+            logging.getLogger("i3pystatus.core.command").error(msg)
+            raise AttributeError(msg)
         command = ["i3-msg", "exec", command]
     else:
-        if not isinstance(command, list):
+        if isinstance(command, str):
             command = shlex.split(command)
 
     try:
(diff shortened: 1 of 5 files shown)


teto:
Cool !

i3pystatus/core/modules.py
```diff
@@ -106,7 +107,7 @@ def split_callback_and_args(cb):
             if cb is not "run":
                 getattr(self, cb)(*args)
         else:
-            run_through_shell(cb, *args)
+            execute(cb, *args)
```
enkore: We probably want detach=True here and no *args - opinions?

richese:
@enkore 
You need the args depending on how you defined the clickevent.
`on_leftclick = "urxvt -e htop"`  ->  `cb = "urxvt -e htop" args = []`
`on_leftclick = ["urxvt", "-e", "htop"]`  ->  `cb = "urxvt" args = ["-e", "htop"]`
The first variant should be sufficient (since you can always create a script for anything more complex) so I'll remove `*args` and update the docs to use only first variant for external commands.

I also realized that 'detached' does not work with parameters because of how i3-msg works so I'll need to look into that too.

richese:
Everything that needed fixing should be good now. :)

I have rewritten a huge chunk of the Callbacks documentation so tell me if you like it this way, see any typos or other grammatical mistakes on my part.

i3pystatus/core/command.py
```diff
@@ -7,13 +7,26 @@
...
+
+    :param command: A string or a list of strings containing the name and
+     arguments of the program.
+    :param enable_shell: If set ot `True` users default shell will be invoked
+     and given ``command`` to execute. The ``command`` should obviously be a
+     string since shell does all the parsing.
     """
 
+    if not enable_shell and not isinstance(command, list):
+        command = command.split()
```
rscholer: Please use `shlex.split(command)` this ensures correct splitting for commands like:
`binary --long-option="more than one word"`.
`command.split()` would result in:
`['binary', '--long-option="more', 'than', 'one' 'word"']`,
whereas `shlex.split(command)` correctly results in
`['binary', '--long-option=more than one word']`.

rscholer: You need to use shlex.split, or a command like this won't be correctly split: `app --long="1 2" a b`.

i3pystatus/core/command.py
```diff
@@ -7,13 +7,26 @@
...
+    output is saved in one variable.
+
+    :param command: A string or a list of strings containing the name and
+     arguments of the program.
+    :param enable_shell: If set ot `True` users default shell will be invoked
+     and given ``command`` to execute. The ``command`` should obviously be a
+     string since shell does all the parsing.
     """
 
+    if not enable_shell and not isinstance(command, list):
```
rscholer: You should also test for tuples, or maybe test for strings. Catching the raised AttributeError is also an option.

PR #258, opened 2015-09-25
Yaourt backend for updates script

schroeji:
I wrote a backend for the updates module to support the AUR helper Yaourt.
The backend supports 2 modes: 
In the aur_only mode it will count only those packages with new updates available from the AUR. So it can be used in conjunction with the pacman backend to get the total update count like this:

    from i3pystatus.updates import pacman, yaourt
    status.register("updates",
        backends = [pacman.Pacman(), yaourt.Yaourt(True)])
This mode can also be disabled to get the total update count directly without using the pacman backend:

    from i3pystatus.updates import yaourt
    status.register("updates",
        backends = [yaourt.Yaourt(False)])

### DIFF
--- a/i3pystatus/updates/yaourt.py
+++ b/i3pystatus/updates/yaourt.py
@@ -1,18 +1,25 @@
 """
 This module counts the available updates using yaourt.
-You should not use it together with the pacman backend otherwise
-some packeges might be counted twice.
+By default it will only count aur packages.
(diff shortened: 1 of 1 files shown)


richese:
You'll need to move the docstring to the module itself like in the other backends to make it show in the documentation.

Also adding the examples from PR to the docstring may be a good idea.

PR #262, opened 2015-10-08
shell: added possibility to set markup language used to parse messages

jpodeszwik:
This commit ads possibility to use pango markup langage to parse messages returned by shell commands, for example:
config:
```
status.register("shell",
    command="<path_to_script>",
    markup="pango",)
```

with this script:
```
#!/bin/bash

echo '<span foreground="red">red text</span> <span foreground="green">green text</span>'
```

would produce something like this:
pango_example

### DIFF
--- a/i3pystatus/shell.py
+++ b/i3pystatus/shell.py
@@ -14,11 +14,13 @@ class Shell(IntervalModule):
 
     color = "#FFFFFF"
     error_color = "#FF0000"
+    markup = "none"
 
     settings = (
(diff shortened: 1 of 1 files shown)


richese:
This is already implemented for all modules.
Just add the
``` python
hints = {"markup": "pango"},
```
setting when registering a module.

http://docs.enkore.de/i3pystatus/configuration.html#hints

jpodeszwik:
You're right. Thanx!

PR #267, opened 2015-10-17
dota2wins: truncate win percentage

drwahl:
Use only 2 decimals for win percentage so we don't fill all of the
status bar with decimal places.

### DIFF
--- a/i3pystatus/core/modules.py
+++ b/i3pystatus/core/modules.py
@@ -94,25 +94,28 @@ class Module(SettingsBase):
         else:
             args = []
 
-        our_method = is_method_of(cb, self)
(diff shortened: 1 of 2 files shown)


enkore:
You can achieve the same without changing the module by formatting. E.g. `format="my dota 2 stats: {win_percentage:0.2f} %"`

drwahl:
Ya, I am aware it can be overridden by passing 'format' when instantiating the module in the config file. I thought it might be good to provide a cleaner default instead of having the default spew a super long percentage number.  I'm happy to keep my local config forcing a 2 decimal percentage, if we don't want to update this default.

PR #268, opened 2015-10-17
plexstatus: initial commit

drwahl:
The plexstatus plugin provides the user with a notification when there
is media being streamed from their Plex Media Server (https://plex.tv/).
The user must provide their API key, which can be found by following the
instructions provided by plex for developers
(https://support.plex.tv/hc/en-us/articles/204059436-Finding-your-account-token-X-Plex-Token).

### DIFF
--- a/i3pystatus/plexstatus.py
+++ b/i3pystatus/plexstatus.py
@@ -1,0 +1,62 @@
+import xml.etree.ElementTree as ET
+from i3pystatus import IntervalModule
+from urllib.request import urlopen
+
+
+class Plexstatus(IntervalModule):
+    """
+    Displays what is currently being streamed from your Plex Media Server.
+    """
+
+    settings = (
+        ("apikey", "Your Plex API authentication key "
+         "(https://support.plex.tv/hc/en-us/articles/204059436-Finding-your-"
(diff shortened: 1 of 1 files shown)


i3pystatus/plexstatus.py
```diff
@@ -0,0 +1,62 @@
...
+        response = urlopen(PMS_STATUS_URL)
+        xml_response = response.read()
+        tree = ET.fromstring(xml_response)
+
+        cdict = {'title': '',
+                 'platform': ''}
+
+        for vid in tree.iter('Video'):
+            try:
+                cdict['title'] = vid.attrib['title']
```
enkore: `cdict["title"] = vid.attrib.get("title", None)` should have the same effect, unless not every `vid` has an `attrib`.

Arvedui:
even if not, that should imho be checked with hasattr not with try except and certainly not with catch all

drwahl:
Oops.  Meant to cleanup that catchall except.  Good catch!  Unfortunately, thanks to the not-so-fantastic APIs, 'vid' and 'plat' may be a null string instead of a legit XML element (in which case, it won't have .attrib).  Patch incoming...

PR #270, opened 2015-10-25
Update desktop.py

janoliver:
Fixed syntax error.


enkore:
Seems I never get the "multi-except" right on the first try! :-)

PR #275, opened 2015-11-09
Added double click support

npcardoso:
This pull request adds support for double mouse clicks in modules.

The improvement is backward-compatible, so plug-ins not using double clicks should not be affected in any way.


kneitinger:
This is great, I really like this idea and can see many uses for it. I can't say I've ever wanted double upscroll or downscroll support, but hey, why not :D

enkore:
merged in https://github.com/enkore/i3pystatus/commit/87c01278f7d6219d14c29f2d2231b280ca77619b

PR #276, opened 2015-11-11
update symbol

ismaelpuerto:


enkore:
U+2328 is a standard Unicode character (keyboard), it should be present in at least some fonts, which makes it a reasonable default in my opinion.

U+F11C is a "private use area" Unicode character, so it isn't really assigned. I guess this comes from an icon font?

PR #277, opened 2015-11-19
 Improved handling of (un)mounted drives/partitions.

douedd:
If a disk or partition is not mounted the current behaviour is to spit an ill-formatted error. This try/except addition silences these and as such will only display information on mounted disks.

### DIFF
--- a/i3pystatus/disk.py
+++ b/i3pystatus/disk.py
@@ -31,10 +31,14 @@ class Disk(IntervalModule):
     round_size = 2
 
     def run(self):
-        stat = os.statvfs(self.path)
-        available = (stat.f_bsize * stat.f_bavail) / self.divisor
+        try:
+            stat = os.statvfs(self.path)
+            available = (stat.f_bsize * stat.f_bavail) / self.divisor
 
-        if available > self.display_limit:
+            if available > self.display_limit:
+                self.output = {}
+                return
+        except:
             self.output = {}
             return
 


Arvedui:
I think it would be better to implement this with a second format string.
Because some people might want to know if a partition is not mounted.

and i think it should default to something like "{path} is not mounted" or something else that is clear and understandable.

Oh and please do not use a wild card except clause.

douedd:
The lack of a message noting that a partition is unmounted is in keeping with the current behaviour for disks whose availability exceeds display_limit. In my opinion the absence of the disk from the statusbar is evidence enough that it is not mounted.

As for the wildcard except, the error thrown in this case is inconsequential -- it's going to return an empty dictionary regardless of why it has failed.


I'll maybe have a look later this evening at a different means of implementing the same behaviour, however I think there are at least some who would approve of the current changes.

richese:
Now that you mention it, I remember doing something like this a half year ago... :cold_sweat: 
Welp, here 3e00099f35ce9e17b725cc1ff7ffb22d9e36affc forgotten in my python folder.

I am in favor of new format setting (possibly defaulting to empty string) as it seems more flexible.

You can try using `os.path.ismount` to check if fs is mounted. That way there is no need for the try catch block.

douedd:
Hm, the ismount() method only seems to be working for the actual base mount point. Any nested dirs are not recognised as mounted.

For example, an sshfs mount at /mnt/sshfs returns True when run against ismount(), however, /mnt/sshfs/home/username returns False.

Using richese's wider code with a try/except is working -- I'll see about the format string.

douedd:
Here is working module with format strings from richese and a modified try/except from before: https://github.com/douedd/i3pystatus/commit/394dbc9f90add11a31fc76ba7e740f64d65cf834

richese:
@douedd 
Well, why would you use a path different from the mountpoint itself in the first place? You'll get the same values either way as it queries data from the disk not folders.

douedd:
For cases where you need to mount the whole filesystem but only care about the free space on the mounted /home partition.

Using the base mountpoint only gives free space from / which is of no concern.

richese:
Then '/home' is a *valid* mountpoint and you'll get free space of '/home' partition.

I am personally using Disk with mountpoints '/', '/var', '/tmp', '/home' localy and '/media/cubie' via NFS and I get the correct values for each of them.

douedd:
Even if that is the case, surely it makes more sense to have a more compatible method (the try/except) for users who would enter the full path, including dirs.

douedd:
Also, in my case the mount points in question are NFS.

So there's a remote filesystem mounted on /mnt/sftp and I only need the free space from /mnt/sftp/home. ismount() only recognises /mnt/sftp as mounted, not /mnt/sftp/home.

richese:
@douedd 
> So there's a remote filesystem mounted on /mnt/sftp and I only need the free space from /mnt/sftp/home. ismount() only recognises /mnt/sftp as mounted, not /mnt/sftp/home.

I see what you meant now. You mount the root '/' of the remote device to '/mnt/sftp' which has '/home' mounted on it. Nice to now, that it works like that too.

>Even if that is the case, surely it makes more sense to have a more compatible method (the try/except) for users who would enter the full path, including dirs.

Yeah, that seems the better point of view.

I just tested the commit you posted but it does not behave like I would expect. 
I guess that the exception is thrown when the path does not exist. That is understandable since '/mnt/sftp/home' can not exist when '/mnt/sftp' is not mounted and that is fine.
But what if I try to get free space on '/mnt/sftp' which is not mounted but the path exists in my system permanently (like there always is /mnt or maybe /media/cdrom even when there is nothing mounted in them). I that case I will get free space of the underlying mountpoint which happens to be '/', without any indication that the filesystem I am actually interested in is not mounted. 

That is the reason why I used 'ismount' and only the mountpoints. Our solutions does not seem to be very compatibile with each others usecases. :smile:

douedd:
Do you think an os.listdir() would be a good solution?

ie. If there are 0 files/directories then assume it's not mounted.


I realise this is far from ideal, but at least if there are no files or directories then you know you have 100% free space.

Will give it a think and get back to you.

richese:
I am fine with using listdir. It lists hidden files so there should be no problem with your reasoning.

I can't think of better solution. If it's not mounted it is just a directory and there is no way of telling whether it is supposed to be used as mountpoint or not.

douedd:
Okay, here's some code with the above implemented: https://github.com/douedd/i3pystatus/commit/01f798e3cfc1c4192e0f53441be15c05cc3732a3

Gave it a whirl and it behaves as expected. Let me know how you get on.

richese:
It works as expected on my end too. :+1:

But the PR seems to be broken. It says it wants to merge 1 commit from unknown repository.

douedd:
Hm, not sure how to fix it without creating another request.

douedd:
I've contacted support in hope that they can correct this. A second request would clutter things up. Hopefully they respond soon.

douedd:
Okay, they said a new pull request was required, so here it is: https://github.com/enkore/i3pystatus/pull/279

PR #278, opened 2015-11-25
VK module

w8u:
Monitors new messages amount in VK social network.

### DIFF
--- a/i3pystatus/vk.py
+++ b/i3pystatus/vk.py
@@ -1,0 +1,74 @@
+from i3pystatus import Status, IntervalModule
+from i3pystatus.core.util import internet, require, user_open
+import vk
+
+
+class Vk(IntervalModule):
+    """
+    Display amount of unread messages in VK social network.
+    Creating your own VK API app is highly recommended for your own privacy, though there is a default one provided.
+    Reference vk.com/dev for instructions on creating VK API app.
+    If access_token is not specified, the module will try to open a request page in browser.
+    You will need to manually copy obtained acess token to your config file.
+    Requires the PyPI package `vk`.
+    """
+
+    API_LINK = "https://oauth.vk.com/authorize?client_id={id}&display=page&revoke=1&scope=messages,offline&response_type=token&v=5.40"
+    app_id = 5160484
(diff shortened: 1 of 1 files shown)


i3pystatus/vk.py
```diff
@@ -0,0 +1,74 @@
...
+    interval = 1
+    color = "#ffffff"
+    color_unread = "#ffffff"
+    color_bad = "#ff0000"
+
+    settings = (
+        ("app_id", "Id of your VK API app"),
+        ("access_token", "Your access token. You must have `messages` and `offline` access permissions"),
+        ("token_error", "Message to be shown if there's some problem with your token"),
+        ("color",),
```
richese: This causes errors in sphinx. Settings should be a string or a tuple of two strings.

You should change this to `"color",` or  `("color", "docstring here"),`.

i3pystatus/vk.py
```diff
@@ -0,0 +1,74 @@
...
+        ("color_bad",),
+        ("color_unread",),
+    )
+
+    @require(internet)
+    def token_request(self, func):
+        user_open(self.API_LINK.format(id=self.app_id))
+        self.run = func
+
+    @require(internet)
```
richese: I would suggest to use `@require(internet)` only for `run` and do some checks there. If you happen to have no internet connection when i3pystatus starts `init` will not be called.

w8u: Actually `run` also seems to be never called if there was no internet connection when i3pystatus started, even if it is the only function where `@require(internet)` is used.

richese: Hmm, it seems to work on my end. [It starts printing when connection is established](https://gist.github.com/richese/8b5b2ad0369e6bc5c4b5).

What interval are you using? `run` is called only every `interval` seconds.

w8u: @richese 
For me it somehow doesn't work. It will start printing only if connection was present at the moment i3pystatus was started. Otherwise nothing shows up even if connection is established later.

enkore:
Merged. GitHub, as so often, did not notice.

PR #279, opened 2015-11-27
 Improved handling of (un)mounted drives/partitions -- take 2.

douedd:
Previously the free space of the underlying filesystem would be reported if the path provided was a directory but not a valid mountpoint. This adds a check to first confirm whether a directory is a mountpoint using os.path.ismount(), and if not, then runs an os.listdir() to count the files; empty directories are considered not mounted.

This functionality allows for usage on setups with NFS and will not report free space of underlying filesystem in cases with local mountpoints as path.

(Original pull request went awry -- https://github.com/enkore/i3pystatus/pull/277)

### DIFF
--- a/i3pystatus/disk.py
+++ b/i3pystatus/disk.py
@@ -37,17 +37,25 @@ class Disk(IntervalModule):
     round_size = 2
     mounted_only = False
 
+    def not_mounted(self):
+        if self.mounted_only:
+            self.output = {}
+        else:
+            self.output = {} if not self.format_not_mounted else {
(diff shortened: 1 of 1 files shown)


enkore:
merged (stupid github doesn't catch on) in

https://github.com/enkore/i3pystatus/commit/1c4f94130419cd9cb73b80c2ad51f12ce6516b02
https://github.com/enkore/i3pystatus/commit/731749f0e3fd535c633ed373ab7e17b86e1e8000

Issue #272, opened 2015-11-03
Add previous_song to spotify plugin & some other bugs


k1-hedayati:
* `previous()` method in `Playerctl` works just fine, we can simply add it to module. 
* `Playerctl` returns length and status but `status` and `length` formats don't show anything even though code seems to be fine.
* When Spotify isn't running i3pystatus doesn't show anything but in code it's supposed to print `Not running` in bar and after running Spotify I have to restart i3pystatus to see info.

t3zla:
Currently playerctl's status command does not work with Spotify 1.0.19, so until that gets fixed I would downgrade to spotify 0.9 which should fix all of those problems. I add the previous command to the module right now

k1-hedayati:
:+1:

enkore:
There have been other changes to the spotify module since #282; if any of the bugs still persist please open one new ticket per bug. Thanks!

PR #282, opened 2015-12-05
Added previous song support to the spotify module, activated by scrolling down while over the module in the status bar

t3zla:
Related to #272


t3zla:
Related to  #272

PR #287, opened 2015-12-17
Added Zabbix module

asmikhailov:
I did zabbix interval module which retrieve all current unacknowledged triggers from zabbix server with Zabbix API credential and show it in full or possibly modified string


asmikhailov:
closed by the reason wrong email

PR #288, opened 2015-12-17
Added Zabbix module

asmikhailov:
I did zabbix interval module which retrieve all current unacknowledged triggers from zabbix server with Zabbix API credential and show it in full or possibly modified string


enkore:
merged https://github.com/enkore/i3pystatus/commit/3d22043881a4115301d58b310fe1ff0c633fe7e0

PR #289, opened 2015-12-22
Added bar showing the absolute capacity in the battery module

lorian1333:
You can use the {bar_design} in the format string of the battery module, which will display the same bar as {bar}, but according to the absolute percentage instead of the relative percentage


enkore:
Merged in https://github.com/enkore/i3pystatus/commit/ef58c5a6fa530c9343ca64748a2f35f5f39c7213

PR #290, opened 2015-12-22
disk module: corrected doc string

mxlian:
The documentation described the parameters wrong. Fixed in commit


enkore:
Good catch. Apropos catch, I'll try to catch up on the pending PRs.

PR #294, opened 2015-12-29
New module solaar

cambid:
This modules checks for a given bluetooth-device via the solaar-cli-utility and prints the status and percentage of the battery.


enkore:
merged in https://github.com/enkore/i3pystatus/commit/b6a472e0143fb0de68bace96aede3272180aadb7

PR #295, opened 2016-01-01
Add Syncthing module

rumpelsepp:
This module provides support for Syncthing [1]. Currently this module
only supports showing the up/down status of Syncthing and it is possible
to start/shutdown Syncthing via click events. A few callback functions
for usage with/without systemd are provided as well.

The module is designed in a generic way (st_get(), st_post()), such that
new features could be add very easily.

[1]: https://syncthing.net

### DIFF
--- a/i3pystatus/syncthing.py
+++ b/i3pystatus/syncthing.py
@@ -1,0 +1,121 @@
+import json
+import os.path
+import requests
+from subprocess import call
+from urllib.parse import urljoin
+import xml.etree.ElementTree as ET
+from i3pystatus import IntervalModule
+from i3pystatus.core.util import user_open
+
+
+class Syncthing(IntervalModule):
+    """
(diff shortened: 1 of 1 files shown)


khertan:
+1
Seems to not pass test due to requests dependancy

rumpelsepp:
How am I supposed to fix this? Other modules use requests as well.

richese:
Looks like you also need to add `requests.exceptions` to MOCK_MODULES in docs/conf.py since you are importing from it.

khertan:
Also I think it ll be a good idea to add a setup to check or not self signed certificate request.get([...],verify=false) as default install of syncthing use self signed certificate

rumpelsepp:
Tests pass now; I have added an `verify_ssl` option.

khertan:
Thx

Issue #304, opened 2016-01-14
Callbacks: Passing self to external callbacks?


mxlian:
The call to the user_open is passing self to it. I saw in other modules they write a wrapper to avoid this, i'll patch it like that.

![screenshot from 2016-01-14 10-23-55](https://cloud.githubusercontent.com/assets/2795329/12320543/74529238-baa9-11e5-8363-d992a5f26509.png)

mxlian:
@enkore I saw  the github and reddit modules having the same problem. They solve it with custom member callbacks for each url they want to open. Could make sense to solve this problem somehow globally?

Like: 
* Add a wrapper to user_open to filter 'self' out, and put it available in **core.utils**
* Changing the user_open to get rid of 'self' (seems wrong to do this)
* Any better solution?

enkore:
Does someone use the availability of *self* to external (non member) callbacks? Even if so, it should be kinda rare and a workaround is possible (Status.register returns the instance which can be passed back to the callback manually if necessary)

mxlian:
For me it seems to be the most clean solution

richese:
I find this feature quite useful actually. For example if you want to wrap a callback to be executed only if certain conditions are met or most likely when you just want value of some module variable.

To me it looks like `user_open` is missing the `self` parameter because it was forgotten when the switch to configurable callbacks happened, since its definition is older than that (14c0528be56485696a8563495daa8352254abb91 and 33aba4b0845d3c278fb0887d750f1feae5e04d78 ). Even its docstring states that it is meant to be called from callbacks, not that it is a callback.
I would add the missing parameter to `user_open` and then check all modules that use it.

Also it could be helpful to rename `self` to `module` in non-member callback [documentation](http://docs.enkore.de/i3pystatus/configuration.html#callbacks) to make it less confusing?

mxlian:
@richese should than the user_open remain in *core.util* ?

enkore:
Another solution came to mind. We could normally remove the self argument, but provide a decorator that adds it back again (e.g. stack inspection), in case one needs self for something more involved.

Pro:
- Easy to use for both cases (any Python function can be used, but involved use cases are not made harder than necessary)
- Only small change needed for existing user callbacks

Con:
- Magic decorator

enkore:
Sample implementation for discussion: https://gist.github.com/enkore/90acffa13df5e7ee8ec7

mxlian:
@richese do you know of a module using this?

[...]

PR #305, opened 2016-01-14
Fixes on bitcoin module

mxlian:


enkore:
This should be resolved now by #304, right?

PR #309, opened 2016-01-27
Added `Timer` module.

richese:
There seems to be an issue with sphinx.
If I add `enum` to mock modules I get error with python 3.5 but then older python complains that it can't find it.


enkore:
Looking at the code I don't think you actually use any of the special benefits of `Enum`, a normal class would work, too.

richese:
@enkore 
You're rigth. Thanks for pointing it out.

PR #310, opened 2016-01-28
Fix for callbacks.

richese:
Removed code did what `hasattr` is already doing one line below.
Also caused exception when member callback was called: trying to call `cb`
which is a `str` as function.

The only change needed for #304 was to remove `self` argument when calling python callback.

Edit: ~~Or maybe not?~~

### DIFF
--- a/i3pystatus/core/modules.py
+++ b/i3pystatus/core/modules.py
@@ -8,14 +8,6 @@ from i3pystatus.core.command import execute
 from i3pystatus.core.command import run_through_shell
 
 
-def is_method_of(method, object):
-    """Decide whether ``method`` is contained within the MRO of ``object``."""
-    for cls in inspect.getmro(object.__class__):
-        if method in cls.__dict__.values():
-            return True
-    return False
-
-
 class Module(SettingsBase):
     output = None
     position = 0
@@ -90,12 +82,9 @@ class Module(SettingsBase):
         else:
             args = []
 
-        our_method = is_method_of(cb, self)
-        if callable(cb) and not our_method:
+        if callable(cb):
             self.__log_button_event(button, cb, args, "Python callback")
             cb(*args)
-        elif our_method:
-            cb(self, *args)
         elif hasattr(self, cb):
             if cb is not "run":
                 self.__log_button_event(button, cb, args, "Member callback")
--- a/tests/test_module_clicks.py
+++ b/tests/test_module_clicks.py
@@ -67,17 +67,17 @@ def test_clicks(events, expected):
         def set_action(self, action):
             self._action = action
 
-        on_leftclick = [set_action, "leftclick"]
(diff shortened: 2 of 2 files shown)


enkore:
With the previous behaviour one could reference method callbacks by name, like test_module_clicks.py does (some other modules use this too, I think). This doesn't work with this patch anymore.

The other way around, referencing a method callback by a string should work with the current state, is_method_of checks whether a function object (unbound method) is within the current objects MRO. It should never return True if you pass it a string.

enkore:
I wrote a test in 8124668 to programmatically express what I mean.

If you checkout a new branch from master and revert it to 4c37834df45 (the last commit before anything here was changed) and then cherry-pick 8124668 into that branch (to get these two tests) you'll see how test_callback_handler_function fails, because there is an extraneous (by our definition) self passed to external functions. However, both tests of test_callback_handler_method, i.e. by-object and by-name/string reference to a method, pass.

richese:
Indeed, I see now that it is needed since `self` is not passed anymore but `is_method_of` definitely can return `True` if you pass it a string.

If you run this example code
``` python
import os
from i3pystatus.core.modules import Module, is_method_of


class Pid(Module):
    on_leftclick = "urxvtc -e ranger"
    on_rightclick = "urxvtc -e htop"

    def init(self):
        self.output = {
            "full_text": "{}".format(os.getpid()),
        }

m = Pid()

print(m.on_leftclick)
print(type(m.on_leftclick))
print(is_method_of(m.on_leftclick, m))

m.on_click(1)
```

it results in this:
```
In [10]: run test.py
urxvtc -e ranger
<class 'str'>
True
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/home/luck/git/i3pystatus/test.py in <module>()
     18 print(is_method_of(m.on_leftclick, m))
     19 
---> 20 m.on_click(1)

/home/luck/git/i3pystatus/i3pystatus/core/modules.py in on_click(self, button)
    182                 m_click.set_timer(button, cb)
    183             else:
--> 184                 self.__button_callback_handler(button, cb)
    185 
    186         return True

/home/luck/git/i3pystatus/i3pystatus/core/modules.py in __button_callback_handler(self, button, cb)
     97         elif our_method:
     98             self.__log_button_event(button, cb, args, "Callable member callback")
---> 99             cb(self, *args)
    100         elif hasattr(self, cb):
    101             if cb is not "run":

TypeError: 'str' object is not callable

```

enkore:
I see, yes, that's a bug in is_method_of. The solution was to simple apparently. I'll fix it

enkore:
739c595 committed, should fix this. There are some cases that one can deliberately construct to fool  it, but in practice they shouldn't occur.

richese:
Works like a charm. :+1:

PR #320, opened 2016-02-14
Update PlaintextKeyring example with new import location.

facetoe:


enkore:
Using keyring-8.4, I don't have a keyrings.alt package, but keyring.backend.file.PlaintextKeyring (still?) exists.

facetoe:
That's weird, I no longer have it:

```
>>> import keyring
>>> from keyring.backends.file import PlaintextKeyring
  File "<stdin>", line 1, in <module>
ImportError: No module named 'keyring.backends.file'
>>> keyring.__version__
'8.4'
>>> keyring.__path__
['/usr/lib/python3.5/site-packages/keyring']
```

I had to install https://pypi.python.org/pypi/keyrings.alt to get it back. 

Maybe I could add a try/except block and attempt importing from both locations?

enkore:
Ah, I see the reason know.

On PyPI there is a .tar.gz, and also a wheel:
https://pypi.python.org/packages/3.5/k/keyring/keyring-8.4-py2.py3-none-any.whl#md5=014f3c41c38f31ad035be909bf370ee4
https://pypi.python.org/packages/source/k/keyring/keyring-8.4.tar.gz#md5=c4c29fcd1d63bf9e299c524ba4fdc04a

The .tar.gz does not contain these backends anymore, as does the GitHub repository. The wheel incorrectly does. I'll report this upstream.

facetoe:
Ahh good catch!

teto:
I noticed I had some problem updating keyrings too.
I use vdirsyncer and they decided to drop native support for keyring and
instead let people run a command, which can be "keyring" (the new
python-keyring programs install a binary called "keyring").
The whole keyring stuff on linux is quite messy :s

2016-02-14 12:51 GMT+01:00 facetoe <notifications@github.com>:

PR #328, opened 2016-03-09
nvidia_temp: add module to display GPU temp

drwahl:
This patch includes a module that displays the GPU tempurate for an
NVidia video card.  The module relies on a system utility called
'nvidia-smi', which is usually installed on the system when the video
card driver is installed.


enkore:
A similar module is already shipped with i3pystatus: http://docs.enkore.de/i3pystatus/i3pystatus.html#gpu-temp

drwahl:
Doh! How did I miss that!? :)

PR #331, opened 2016-03-14
Add wunderground module and implement generic weather module

terminalmage:
**NOTE: Since submitting this pull request, major changes have been made. A generic weather module has been added, and the weather.com and wunderground modules have been reworked as backends for the generic module.**

This module tries to use as much of the same variable naming conventions
that the ``weather`` module uses as possible, to make transitioning
easier in the future in case we decide to make a base class for all
modules which provide weather data.

An API key is required to use this module, information on obtaining one
can be found in the docstring.


enkore:
Hm, wundergrund and weather are very similar indeed. Maybe refactor the weather module into a module and a weathercom backend class, and then add wunderground as another backend?

`status.register("weather", backend="wunderground", api_key="…"…)`

=> `def init(…): self.backend = Weather.backends[self.backend](self.api_key)`

terminalmage:
Yeah, I thought this might be necessary. I'll work on this over the next few days and update the pull request.

terminalmage:
@enkore I'm still working on this, but I noticed something odd. On two occasions now I've gotten API usage alerts for making more than 10 API requests in a minute. This should never happen since I have the interval set to an extremely conservative 1800 seconds. The only explanation I can think of is that, since both API usage alerts have coincided with my resuming my laptop from sleep, that i3pystatus is realizing that my laptop has resumed from suspend after a number of hours, and executes the check once for each time the interval elapsed in the time between the laptop was put to sleep and when it was resumed.

This either seems like a bug i3pystatus.core, or a bug in how I wrote the wunderground plugin. As I used the weather.com plugin as a guide, I don't know for sure if I did anything obviously wrong.

enkore:
It uses wall clock time to track time, so if your interval is, say, 30 minutes, and goes to sleep for an hour, then i3pystatus will run it immediately and not way another 30 minutes.

It shouldn't run it more than once. I'll look into it.

terminalmage:
@enkore I have completed the rework of both the weather.com and wunderground modules as backends for a newly-added generic weather module.

This is a pretty substantial change and adds a bunch of new formatters for weather.com, in addition to supporting ``formatp()`` syntax, as well as adding ``on_leftclick`` support to launch the weather.com forecast page (this had previously been included in my new wunderground module, and is now implemented in the generic weather module).

I tested my RST by building the docs using sphinx, and the links and references all appear to work and non appear to be broken.

This will cause non-compatible changes when people update, so it should probably be mentioned in the release notes for the next tagged release if i3pystatus.

terminalmage:
Github appears to be including a commit that was squashed into a different commit and then force-pushed, so it shouldn't be showing here. To avoid potential issues, I will close this and open a new pull request.

PR #333, opened 2016-03-17
Fix KeyError when using backend-specific updates formatter

terminalmage:
When a backend-specific formatter (i.e. ``{Pacman}``, ``{Cower}``, etc.)
is used, and the initial "working" status is set, the loop in which the
update totals is compiled has not yet run, leading to a KeyError.

This commit fixes the traceback by setting initial values of "?" for
these formatters before the initial "working" status is set.

### DIFF
--- a/i3pystatus/updates/__init__.py
+++ b/i3pystatus/updates/__init__.py
@@ -92,6 +92,11 @@ class Updates(Module):
 
     @require(internet)
     def check_updates(self):
+        for backend in self.backends:
+            key = backend.__class__.__name__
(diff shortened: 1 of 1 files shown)


terminalmage:
Actually, I need to rethink this... the call to formatp is apparently there to signify that check_updates() is compiling the count. In this case it's probably best to set these keys in ``self.data`` to empty strings or something like that.

terminalmage:
OK, I've redone this and edited the body of the pull request message above.

enkore:
Good catch

PR #334, opened 2016-03-18
lock: A simple module to lock the screen

drwahl:
When the user clicks on the lock icon, a lock screen command will run,
locking the users workstation.

### DIFF
--- a/i3pystatus/lock.py
+++ b/i3pystatus/lock.py
@@ -1,0 +1,31 @@
+from subprocess import getoutput
+from i3pystatus import IntervalModule
+
+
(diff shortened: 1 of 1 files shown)


enkore:
This should be possible with the [Text](https://docs.enkore.de/i3pystatus/i3pystatus.html#text) module, something like this

```
status.register("text",
on_leftclick="i3lock",  # or something
text="Lock the screen",)
```

drwahl:
Ah, was unaware of that the text module had 'on_leftclick' support. Thanks for pointing this out!

Issue #335, opened 2016-03-21
solaar shows an error if the mouse if off or out of battery


hobarrera:
Whenever I turn the mouse off (eg: because I'll be AFK, or not really using it much), the solaar plugin displays (in red!):

    problem finding battery status device Performance Mouse MX

solaar should remember the mouse, and merely say it's offline, rather than show a huge message (which overflows i3pystatus from the screen).

PR #337, opened 2016-03-21
Improve solaar error handling/display

hobarrera:
Improve errors shown by solaar plugin, also cleaning up how error are internally handled.

See #335


hobarrera:
This actually fixes #335, since the other error is due to a bug in Solaar (which already has an unreleased fix).

Issue #259, opened 2015-09-27
[question] on_middleclick callback


Anon1234:
Can't find anything about middle click callback in docs. Is there a way to bind an action to it?

teto:
From modules.py
```
   6     settings = (                                                                                 
    27         ('on_leftclick', "Callback called on left click (see :ref:`callbacks`)"),                
    28         ('on_rightclick', "Callback called on right click (see :ref:`callbacks`)"),              
    29         ('on_upscroll', "Callback called on scrolling up (see :ref:`callbacks`)"),               
    30         ('on_downscroll', "Callback called on scrolling down (see :ref:`callbacks`)"),           
    31         ('on_doubleleftclick', "Callback called on double left click (see :ref:`callbacks`)"),   
    32         ('on_doublerightclick', "Callback called on double right click (see :ref:`callbacks`)"), 
    33         ('on_doubleupscroll', "Callback called on double scroll up (see :ref:`callbacks`)"),     
    34         ('on_doubledownscroll', "Callback called on double scroll down (see :ref:`callbacks`)"), 
    35         ('multi_click_timeout', "Time (in seconds) before a single click is executed."),         
    36         ('hints', "Additional output blocks for module output (see :ref:`hints`)"),              
    37     )                    

```
so no it's not supported.
Clicks could use an update to support buttons based on the click number (and keep aliases) for backwards compatibility. I've been thinking about this for quite some time without the time to implement it. I would be glad to use previous/next buttons of mouse.

enkore:
Middle mouse button could be added as usual, and we could add something like `on_unhandled` or `on_fallback` that gets all click events that were not handled otherwise.

hasB4K:
I have, along with a better click support, do like you said @enkore:
* Simplify the switch between all action
* Add the middle click support
* In case of failing, we can handle that in a unhandled callback

If you have any suggestion on this, I will be glad to hear it.

teto:
We should support all mouseclicks, i.e. if a mouse has 30 buttons, i3pystatus should support it. I know I would like to be able to use my previous/next mouse buttons. i3 sends the button id, so upon click, i3pystatus could check if there is an on_click<ID> parameter and use that as a callback. For backwards compatibility, one can provide a dict to map on_leftclick to on_click0, on_rightclick to on_click2 (IIRC) etc...

hasB4K:
I have added the possibility with #339 to have the mouse positions and the button_id in a callback.
[...]

PR #339, opened 2016-03-23
core: Change command_endpoint and on_click for supporting i3bar mouse positions

hasB4K:
Hello,

Here the change of this patch:

* `command_endpoint`: get the position from the mouse when the click
  occured. Parameters names are set here: `pos_x` `pos_y`.
  Positions are passed to on_click through keyword arguments.
* `Module`:
  - change `__log_button_event`, `__button_callback_handler` and `on_click`
    methods for handling keyword arguments.
  - "Member", "Method" and "Python" callbacks are handled by detecting
    if they have `pos_x` or `pos_y` as parameters, or if they have a
    keyword arguments. The special case of wrapped callbacks (made with
    `get_module` decorator for example) is handled in a similar way.
  - "External command" is handled by considering the position as a
    format dictionary. Actually no distinctions are made of how
    `self.data` and the new keyword argument are treated on this.
  - the parameter `kwargs` as been added to the doc string of `on_click`.
* `MultiClickHandler`: now handle keyword arguments.

It's a great feature for adding graphical interface on the top of the status bar.
I think we should add this to the doc if the change is accepted :)

Regards,

-- 
Mathis 'hasB4K' FELARDOS

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -39,6 +39,7 @@ Kenneth Lyons
 krypt-n
 Lukáš Mandák
 Łukasz Jędrzejewski
+Mathis Felardos
 Matthias Pronk
 Matthieu Coudron
 Matus Telgarsky
--- a/i3pystatus/core/__init__.py
+++ b/i3pystatus/core/__init__.py
@@ -29,9 +29,16 @@ class CommandEndpoint:
         self.thread.start()
 
     def _command_endpoint(self):
-        for command in self.io_handler_factory().read():
-            target_module = self.modules.get(command["instance"])
-            if target_module and target_module.on_click(command["button"]):
+        for cmd in self.io_handler_factory().read():
+            target_module = self.modules.get(cmd["instance"])
+
+            try:
+                pos = {"pos_x": int(cmd["x"]),
+                       "pos_y": int(cmd["y"])}
+            except Exception:
+                pos = {}
+
+            if target_module and target_module.on_click(cmd["button"], **pos):
                 target_module.run()
                 self.io.async_refresh()
 
--- a/i3pystatus/core/__init__.py
+++ b/i3pystatus/core/__init__.py
@@ -32,14 +32,16 @@ class CommandEndpoint:
         for cmd in self.io_handler_factory().read():
             target_module = self.modules.get(cmd["instance"])
 
+            button = cmd["button"]
+            kwargs = {"button_id": button}
             try:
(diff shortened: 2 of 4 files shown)


hasB4K:
All right, it seems there is an issue only on Python 3.3. The tests were OK on Python 3.5 and Python 3.5...
I know where this happening, and I'm trying to fix this.

hasB4K:
So, the problem were coming from the different implementations of `inspect` between Python 3.3 and Python 3.4. Indeed on python3.4 the function `inspect.getargspec` handle Functors.

Here an example:

file: test-functor.py
```python
import inspect

class Functor_test:
    def __call__(self, test, test1=42, **kwargs):
        print(test)

print(inspect.getargspec(Functor_test()))
```

Output on the different version of Python:
```
$ python3.3 test-functor.py
  File "test-functor.py", line 7, in <module>
    print(inspect.getargspec(Functor_test()))
  File "/usr/lib/python3.3/inspect.py", line 824, in getargspec
    getfullargspec(func)
  File "/usr/lib/python3.3/inspect.py", line 851, in getfullargspec
    raise TypeError('{!r} is not a Python function'.format(func))
TypeError: <__main__.Functor_test object at 0x7f6f73988dd0> is not a Python function

$ python3.4 test-functor.py
ArgSpec(args=['self', 'test', 'test1'], varargs=None, keywords='kwargs', defaults=(42,))

$ python3.5 test-functor.py
ArgSpec(args=['self', 'test', 'test1'], varargs=None, keywords='kwargs', defaults=(42,))
```

facetoe:
I really like this idea, it would be awesome to be able to add graphical elements to i3pystatus modules!

hasB4K:
@facetoe Yes indeed :)
I have already done that in modifying (duck typing) `command_enpoint` in my configs file, then I can call a Gtk Calendar with the proper x and y positions.

Of course there are still some limitations to this:
* You need to calculate an offset before launching anything (in case if you are on the border of the screen)
* When you change, with `xrandr`, the number of screens or their position/resolution, i3 doesn't update properly, and the x and y positions are wrong (I think I should open an issues on this!). You are obligated to do a: `i3-msg restart`

hasB4K:
Because this pull request modify the behavior of on_click, I thought it was a good idea to fix the issue #259 along the way :)

enkore:
Nice!

teto:
thanks for the great patch, it adds a realm of possibilities !

facetoe:
Awesome patch, can't wait to experiment with it!

PR #340, opened 2016-03-23
Pulseaudio: Display/control active sink

grimpy:
Make sink a property which checks which sink is currently active.
Use pactl to control volumes which gets standard insalled with libpulse
which is already a requirement.


enkore:
That's much nicer than calling amixer to modify PulseAudio state

PR #344, opened 2016-03-25
Respect keyring_backend settings

stratosgear:
Uses the keyring_backend keyring to get the Github password.  For some reason this was not properly used.

Also adds an extra option to redirect the user to their Github profile page (the default option as it was before)  or directly to their notifications page.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -39,8 +39,8 @@ MOCK_MODULES = [
     "dateutil",
     "httplib2",
     "oauth2client",
-    "apiclient"
-
+    "apiclient",
+    "keyring"
 ]
 
 for mod_name in MOCK_MODULES:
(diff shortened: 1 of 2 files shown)


facetoe:
The most likely reason that the Github module did not retrieve your credentials is you did not set them using the `setting_util.py` program provided withi3pystatus. This is necessary as the code that retrieves these credentials uses a specific naming convention: `package.module.Class.setting`. 

When the credentials are properly set they are retrieved and assigned during the module loading process: https://github.com/enkore/i3pystatus/blob/master/i3pystatus/core/settings.py#L109.

stratosgear:
Wow, didn't even know this util existed.  Kinda weird way going about it though...  Let me see what I can do with that!

enkore:
The keyrings stuff basically expects that the application using it has some sort of direct UI to enter credentials, but i3pystatus has no such thing. Details in the [docs](https://docs.enkore.de/i3pystatus/configuration.html#credentials).

PR #345, opened 2016-03-27
Fix sink selection bug.

facetoe:
A sink in the SUSPENDED state would never be selected. Not sure if this
solution is correct...

### DIFF
--- a/i3pystatus/pulseaudio/__init__.py
+++ b/i3pystatus/pulseaudio/__init__.py
@@ -95,10 +95,11 @@ class PulseAudio(Module, ColorRangeModule):
         state = b'DEFAULT'
         for sink in sinks.splitlines():
             attribs = sink.split()
-            if attribs[-1] == b'RUNNING':
+            sink_state = attribs[-1]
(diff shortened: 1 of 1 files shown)


i3pystatus/pulseaudio/__init__.py
```diff
@@ -95,10 +95,11 @@ def sink(self):
         state = b'DEFAULT'
         for sink in sinks.splitlines():
             attribs = sink.split()
-            if attribs[-1] == b'RUNNING':
+            sink_state = attribs[-1]
+            if sink_state == b'RUNNING':
                 bestsink = attribs[1]
                 state = 'RUNNING'
-            elif attribs[-1] == b'IDLE' and state == b'DEFAULT':
+            elif (sink_state == b'IDLE' or sink_state == b'SUSPENDED') and state == b'DEFAULT':
```
enkore: `sink_state in [b'IDLE', b'SUS...']` ?

facetoe: Yes that's nicer, I'll change it.

PR #358, opened 2016-04-03
Now playing: be more tolerant for mpris properties

grimpy:
Some mpris clients dont implement all properties

### DIFF
--- a/i3pystatus/now_playing.py
+++ b/i3pystatus/now_playing.py
@@ -100,14 +100,20 @@ class NowPlaying(IntervalModule):
         try:
             player = self.get_player()
             properties = dbus.Interface(player, "org.freedesktop.DBus.Properties")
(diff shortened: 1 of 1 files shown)


i3pystatus/now_playing.py
```diff
@@ -100,14 +100,20 @@ def run(self):
         try:
             player = self.get_player()
             properties = dbus.Interface(player, "org.freedesktop.DBus.Properties")
-            get_prop = functools.partial(properties.Get, "org.mpris.MediaPlayer2.Player")
+
+            def get_prop(name, default=None):
+                try:
+                    return properties.Get("org.mpris.MediaPlayer2.Player", name)
+                except:
```
enkore: Can you make this `except` clause less broad? It probably raises AttributeError...

enkore:
Nice, thanks!

PR #361, opened 2016-04-07
Introduce latency plugin (IPv4 only at the moment)

Xenopathic:
This plugin runs `ping` at regular intervals against a host on the internet to test the current internet connection latency. I built this plugin after using the 'online' plugin, but wanting a way to see the actual connection latency instead of just online/offline status.

The external `ping` utility is necessary since sending ICMP messages requires special capabilities, and it would be impractical to require i3pystatus to be installed with these capabilities which require root privileges to set. Also, there's no Python way to send ICMP messages.


enkore:
#348 ?

Xenopathic:
@enkore Wow, I didn't even see that other PR (I'm running 3.34 so didn't notice the new code). Awesome that it's already implemented!

PR #362, opened 2016-04-07
RFC: Improved logging

terminalmage:
This pull request adds a new parameter to the ``i3pystatus.status.Status()`` class
called ``logformat``, which allows the user to set the log format. It also
establishes a default logformat.

@enkore I'd like to get your opinion on this. Maybe the default logformat I
chose should be different? Also, we'd probably want to add this information to
the documentation (I have yet to do this). **EDIT:** I have now added documentation.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -233,7 +233,23 @@ to files in your home directory named
 ``.i3pystatus-<pid-of-thread>``. Some modules might log additional
 information.
 
-.. rubric:: Log level
+Setting a specific logfile
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When instantiating your ``Status`` object, the path to a log file can be
+specified. If this is done, then log messages will be sent to that file and not
+to an ``.i3pystatus-<pid-of-thread>`` file in your home directory.  This is
+useful in that it helps keep your home directory from becoming cluttered with
+files containing errors.
+
+.. code-block:: python
+
+    from i3pystatus import Status
+
+    status = Status(logfile='/home/username/var/i3pystatus.log')
+
+Log level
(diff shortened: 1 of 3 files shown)


terminalmage:
I added documentation of the proposed new option. I also added documentation for the ``logfile`` param in ``i3pystatus.status.Status()``, and did it in a separate commit to keep the changes atomic. That way, if this pull request gets rejected, I can still take f40eb15 and submit it in a new PR to improve documentation on the already-existing ``logfile`` param.

enkore:
I think the log format is fine, the most important thing (full traceback) will be visible with any format.

terminalmage:
The configuration (logfile, format, etc.) is passed through to child loggers, even ones created within backend instances for the things like updates, mail, and weather.

Just from a development standpoint, I find it immensely useful to have more information in the log entry. For instance, right now I am working on a new multi-backend module which uses the logger very heavily, at different loglevels (with the vast majority of it in debug), and having name of the loglevel at minimum would be very useful.

Perhaps a better default log format would be ``%(asctime)s %(levelname)s:`` ?

terminalmage:
So, right now I am trying to troubleshoot a module that does all of its updates in a separate thread, which is not reliably resuming its checks after my laptop resumes from sleep. I have a lot of log information, but no timestamp for any of it.

terminalmage:
This is the sort of problem I am trying to solve with this PR.

terminalmage:
@enkore Thanks for merging. Please change the default log format if you prefer, I think the one I used in my PR was overly verbose. Something like I suggested earlier (``%(asctime)s %(levelname)s:``) is much more concise and should suffice for the majority of users.

Again, thanks. This will be a TON of help.

Issue #359, opened 2016-04-06
Bad status line in xkblayout


k1-hedayati:
When I run `subprocess.check_output("setxkbmap -query | awk '/layout/,/variant/{print $2}'", shell=True).decode('utf-8').replace("\n", " ").strip()` this is what I get `us,ir grp:menu_toggle,caps:hyper,grp_led:scroll` which I don't think is a good output, layout check command needs refactoring.

enkore:
#365

PR #365, opened 2016-04-10
Stricter criteria for including query output

ncoop:
Addresses #359. 

The prior code's awk mistakenly assumed a `variant` line always follows a `layout` line, resulting in all following lines being included.

However, user should get a `format` option if they want to include this stuff, or at least an option to not have it uppercased.


ncoop:
Is it better to rely on awk, or make python do the line processing?

Issue #364, opened 2016-04-09
Display actual updates


ncoop:
The updates module uses whatever backend is available to get the number of updates. The output of the check operation should be preserved so the user can see it.

There's already DesktopNotification in desktop.py, and icon sets include `software-update-available`. It's straightforward to set up a call to it in Updates, perhaps `on_rightclick`. It's not the only way to display though.

However, there needs to be a field set up in updates to store the text output, and all of the backends need to be updated to use it. How to deal with multiple backends being allowed?

enkore:
> How to deal with multiple backends being allowed?

Unique package names, if multiple backends supply a package name, the later backend wins?

enkore:
Implemented by #366

PR #366, opened 2016-04-10
Notification with update summary on_rightclick

ncoop:
Working for DNF, re. #364.

The other backends should be updated to return, as the second element of a tuple, what they want in the notification body. (I'm not prepared to do them myself.) With no changes, the notif balloon should be OK but will only have the count.

updates-dnf

### DIFF
--- a/i3pystatus/updates/__init__.py
+++ b/i3pystatus/updates/__init__.py
@@ -2,6 +2,7 @@ import threading
 
 from i3pystatus import SettingsBase, Module, formatp
 from i3pystatus.core.util import internet, require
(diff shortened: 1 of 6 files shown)


ncoop:
Added three commits to PR.

facetoe:
This is cool, I've tested the Cower backend and it works great.

enkore:
Works here nicely with pacman+cower as well, very nice work

ncoop:
Thanks. Note that the `updates` method of any backend must return an int and string.

Also, I haven't tested `aptget` myself.

PR #376, opened 2016-05-01
added dynamic color support for cpu_usage

scattenlaeufer:


i3pystatus/cpu_usage.py
```diff
@@ -29,20 +32,39 @@ class CpuUsage(IntervalModule):
...
         self.prev_busy = defaultdict(int)
         self.formatter = Formatter()
 
+        self.key = re.findall('usage_cpu\d+', self.format)
+        if len(self.key) == 1:
+            self.key = self.key[0]
+        else:
+            self.key = 'usage_cpu'
+
+        if not self.color:
```
facetoe: No need to set this explicitly. It's probably best to just not set the color key in the output if no color is specified and dynamic_color is False.

PR #387, opened 2016-05-21
Google calendar modifications

facetoe:


i3pystatus/google_calendar.py
```diff
@@ -5,10 +5,11 @@
 import pytz
 from apiclient import discovery
 from dateutil import parser
+from googleapiclient.errors import HttpError
```
enkore: docs/conf.py MOCK_MODULES

facetoe: Ahh yes good idea. Sorry I forgot to check if the CI build failed or not so didn't see this error, I'll fix up the docs and resubmit.

PR #388, opened 2016-05-21
add weekcal module

eBrnd:
I often find myself opening a console and using `cal` to find out things like "which day is the 15th" or "what date was this tuesday". So much that I wanted to permanently have that info in my i3pystatus. This module displays an entry like this: `Sat  16 17 18 19 20[21]22  May 2016`

### DIFF
--- a/i3pystatus/weekcal.py
+++ b/i3pystatus/weekcal.py
@@ -1,0 +1,70 @@
+from calendar import Calendar
+from datetime import date, timedelta
+
+from i3pystatus import IntervalModule
+
+
+class WeekCal(IntervalModule):
+    """
+    Displays the days of the current week as they would be represented on a calendar sheet,
+    with the current day highlighted.
+    By default, the current day of week is displayed in the front, and the month and year are
+    displayed in the back.
+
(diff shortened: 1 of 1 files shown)


i3pystatus/weekcal.py
```diff
@@ -0,0 +1,70 @@
...
+    displayed in the back.
+
+    Example: ``Sat  16 17 18 19 20[21]22  May 2016``
+    """
+
+    settings = (
+        ("startofweek", "First day of the week (0 = Monday, 6 = Sunday), defaults to 0."),
+        ("prefixformat", "Prefix in strftime-format"),
+        ("suffixformat", "Suffix in strftime-format"),
+        ("todayhighlight", "Characters to highlight today's date"),
```
enkore: Maybe make that a two-tuple to allow e.g. colored output with something like `todayhighlight=('<span color="#0000ff">', '</span>')`

eBrnd: Good idea, will do that

eBrnd:
Turns out the only change needed was the default parameter :grinning:

PR #389, opened 2016-05-22
spotify: pass player_name=spotify to playerctl

enkore:


enkore:
confirmed fix, thanks

PR #393, opened 2016-06-05
shell: add hide_if_empty setting

colajam93:
This PR adds `hide_if_empty` setting to shell module.
This setting hides status code output when the command successful and the output is empty.
It is useful for commands which succeed with empty output, for example [netctl-auto current](https://git.archlinux.org/netctl.git/tree/docs/netctl-auto.1.txt) returns 0 with empty when there is no active profile.


colajam93:
I found that empty `span` element and disabling separator can do same thing, so I close this PR.
Thanks.

PR #400, opened 2016-06-14
Treat i3pystatus as launcher

igsha:
Hi!
I'd like to use i3pystatus python executable as launcher for configuration files:
```bash
i3pystatus -c ~/.config/i3/status-config.py
```
The main reason is that executables in my system (NixOS) has own interpreter: from ld-linux.so for binaries to shabang interpter for scripts. Another reason is that some well-knonw status applications have this functionality: xmobar, conky, i3status. And the last one: I can install i3pystatus inside virtualenv and launch configuration files directly with the help of `i3pystatus -c`.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -123,6 +123,16 @@ Also change your i3wm config to the following:
 .. note:: Don't name your config file ``i3pystatus.py``, as it would
     make ``i3pystatus`` un-importable and lead to errors.
 
+Another way to launch your configuration file is to use ``i3pystatus`` script
+from installation:
+
+.. code:: bash
+
+    i3pystatus -c ~/.path/to/your/config/file.py
+
+If no arguments were provided, ``i3pystatus`` script works as an example of
+``Clock`` module.
+
 .. _credentials:
 
 Credentials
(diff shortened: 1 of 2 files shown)


i3pystatus/__init__.py
```diff
@@ -25,9 +27,23 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(description='''
+        run i3pystatus configuration file. Starts i3pystatus clock example if no arguments were provided
+    ''')
+    parser.add_argument('-c', '--config', help='path to configuration file', default=None, required=False)
+    args = parser.parse_args()
+
+    if args.config is not None:
```
rscholer: This line should be:
`if args.config:`
 it's more pythonic this way.

enkore:
Can you document this in the sphinx docs?

docs/configuration.rst
```diff
@@ -123,6 +123,16 @@ Also change your i3wm config to the following:
 .. note:: Don't name your config file ``i3pystatus.py``, as it would
     make ``i3pystatus`` un-importable and lead to errors.
 
+Another way to launch your configuration file is to use ``i3pystatus`` script
+from intallation:
```
enkore: installation

PR #404, opened 2016-06-19
If imaplib2 is installed use the IDLE feature of IMAP to avoid polling

janoliver:
I sort of copied the code together, without really understanding what goes on in the thread checking for mail. It seems to work, though. 

Ideally, there should be some cleanup such as 

    idler.stop()
    idler.join()
    self.conn.close()
    self.conn.logout()

after the script exits, I just don't know where to put those. Also, I moved the login command to the `init` function. I am not sure if this breaks anything when suspending the computer or temporarily losing connection, as I don't know how i3pystatus handles connection losses. 

Feel free to change stuff so it is more stable.

### DIFF
--- a/i3pystatus/mail/imap.py
+++ b/i3pystatus/mail/imap.py
@@ -1,8 +1,60 @@
-import imaplib
+try:
+    from imaplib2 import IMAP4, IMAP4_SSL
+    use_idle = True
+except:
+    from imaplib import IMAP4, IMAP4_SSL
+    use_idle = False
 import socket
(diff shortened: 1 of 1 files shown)


enkore:
Conflicts with base, please rebase.

janoliver:
I noticed that the module fails when the connection is lost. I'll patch that someday and rebase.

janoliver:
Ok, I added some failure recovery and rebased onto your master.

enkore:
Thanks!

PR #407, opened 2016-06-22
Change default weather icon for thunderstorm, add "Fog" weather condition

terminalmage:
I had added this to be able to use a different icon for thunderstorms,
which were at the time not even being detected as a weather condition
and were just falling back to the default (i.e. no colorization nor
icon). However, the only thunder/lightning unicode icon that is widely
available is too tall, leading to the entire module's text being
vertically-aligned on a different plane from the rest of the modules in
the status bar.

This commit changes the default icon to the same one used for "Rainy"
conditions, while preserving it as a distinct weather type so that
others can still use their own icon if they so choose.

**EDIT: Thanks to @ncoop I changed this PR to include a better "lightning" unicode icon.**

### DIFF
--- a/i3pystatus/weather/__init__.py
+++ b/i3pystatus/weather/__init__.py
@@ -89,7 +89,7 @@ class Weather(IntervalModule):
         'Cloudy': (u'\u2601', '#f8f8ff'),
         'Partly Cloudy': (u'\u2601', '#f8f8ff'),  # \u26c5 is not in many fonts
         'Rainy': (u'\u26c8', '#cbd2c0'),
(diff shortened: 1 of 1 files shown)


terminalmage:
Example:

weather

ncoop:
How about u26A1?
http://www.fileformat.info/info/unicode/char/26a1/index.htm
⚡

terminalmage:
That might work, checking.

terminalmage:
u2681 works well enough to use, thanks for the idea. I updated the icon and also added a condition for "Fog" (which right now just has a color code and no icon, since I haven't found a good icon for it that's widely available).

PR #411, opened 2016-07-17
Pulseaudio: Update status immediately when volume changes

jcornwall:
Volume changes may occur externally (e.g. through pactl bound to media
keys) or through mouse interaction. The new volume is received
asynchronously but can only be outputted at the next status poll,
causing a small delay.

Introduce Module.send_output() to allow a module to send its output
without delay. Invoke this method in the Pulseaudio sink info callback.

### DIFF
--- a/i3pystatus/core/modules.py
+++ b/i3pystatus/core/modules.py
@@ -88,6 +88,10 @@ class Module(SettingsBase):
     def run(self):
         pass
 
+    def send_output(self):
+        """Send a status update with the current module output"""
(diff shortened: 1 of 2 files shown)


ncoop:
Interesting. Can `self.send_output` request a bar update with any module?

jcornwall:
Yes, although this sequence makes sense only for modules which update `self.output` outside of `Module.run`.

facetoe:
Nice work @jcornwall. I've tested this on my setup and it works great.

PR #414, opened 2016-07-23
Allow use of access_token in github module

nclsHart:
Implements an alternative way to authenticate through GitHub API, see https://developer.github.com/v3/#authentication for more informations.

### DIFF
--- a/i3pystatus/github.py
+++ b/i3pystatus/github.py
@@ -7,8 +7,15 @@ from i3pystatus.core.util import user_open, internet, require
 
 class Github(IntervalModule):
     """
-    Check Github for pending notifications.
+    Check GitHub for pending notifications.
     Requires `requests`
+
+    Availables authentication methods:
+
+    * username + password
+    * access_token (manually generate a new token at https://github.com/settings/tokens)
+
+    See https://developer.github.com/v3/#authentication for more informations.
 
     Formatters:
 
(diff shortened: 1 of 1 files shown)


i3pystatus/github.py
```diff
@@ -34,6 +35,7 @@ class Github(IntervalModule):
         ('unread_marker', 'sets the string that the "unread" formatter shows when there are pending notifications'),
         ("username", ""),
         ("password", ""),
+        ("access_token", ""),
```
enkore: Can you put the link to the docs you mentioned here? Looks otherwise good, thanks.

nclsHart:
@enkore done.

I also add a link to the tokens generation page.

enkore:
Nice, thanks!

ncoop:
Having an issue with this, but I don't think it has anything to do with your implementation, @nclsHart
```python
from json import loads
from subprocess import check_output
from requests import get


def report(data):
    if isinstance(data, list):
        print("list length: {}".format(len(data)))
    else:
        print(data)
    print()

full_url = "https://api.github.com/notifications" \
           "?access_token=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"

print("with curl:")
raw = check_output("curl --silent {}".format(full_url), shell=True)
data = loads(raw.decode("utf-8"))
report(data)

print("with requests:")
response = get(full_url)
data = loads(response.text)
report(data)
```

Result:
```
with curl:
list length: 3

with requests:
{'documentation_url': 'https://developer.github.com/v3', 'message': "Missing the 'notifications' scope."}
```

It's the same token, but GitHub won't take it if sent with `requests`?

nclsHart:
I copy/paste your code and I get the following result: 

```
with curl:
list length: 1

with requests:
list length: 1
```

Maybe you should retry with another token.

ncoop:
@nclsHart thanks for testing. It does work for me on other hosts.

PR #416, opened 2016-07-24
Add format option, logic to handle suspended MLB games

terminalmage:
Tonight's game was the first time I've seen the API return from a
suspended game. I added a separate format option for it.

I've also fixed date/time detection. The API returns different fields for if
a game ended after midnight eastern, and also different ones when a game
is suspended.


terminalmage:
This isn't quite ready yet, need to make one more tweak.

terminalmage:
OK, done tweaking.

PR #420, opened 2016-07-30
mpd: implement stop method

ncoop:
`next_song` was previously bound to `on_rightclick`, and is still bound by default to `on_upscroll`.

### DIFF
--- a/i3pystatus/mpd.py
+++ b/i3pystatus/mpd.py
@@ -27,7 +27,15 @@ title is empty)
     * `{bitrate}` — (Current bitrate in kilobit/s)
     * `{volume}` — (Volume set in MPD)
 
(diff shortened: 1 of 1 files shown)


i3pystatus/mpd.py
```diff
@@ -58,8 +58,8 @@ class MPD(IntervalModule):
     truncate_fields = ("title", "album", "artist")
     hide_inactive = False
     on_leftclick = "switch_playpause"
-    on_rightclick = "next_song"
```
facetoe: I think these can be configured in your config, maybe best to leave these as they are for backwards compatibility?

ncoop:
Alright, done and squashed.

PR #421, opened 2016-07-30
Document how to customize the internet check

ncoop:
Also, move Credentials afterward since it's the most specific topic.


facetoe:
Awesome thanks.

PR #422, opened 2016-07-31
Pulseaudio: Wait on child processes to avoid zombies

jcornwall:
subprocess.Popen() does not implicitly wait() on the child process.
This leads to transient zombie pactl processes which are only reaped
as new processes are spawned.

Use synchronous subprocess communication to collect child process
return codes immediately and allow them to exit.


facetoe:
Nice, thanks.

PR #425, opened 2016-08-06
Add module for streaming ABC radio Australia.

facetoe:


facetoe:
Hmm that's a funny error.. Is Sphinx executing with Python2.7 or something?

richese:
`enum` was added in 3.4 
i3pystatus requieres only 3.2

facetoe:
Ahh OK, I'll just make it a regular class as well.

enkore:
I guess we could bump the required Python version for the next release, since 3.4 and 3.5 are pretty much the standard now

Issue #424, opened 2016-08-05
`required` constraint is not enforced


ncoop:
Modules have a `required` set that specifies options that must be configured, or else `i3pystatus.core.exceptions.ConfigMissingError` is thrown.

If a config in that set is _initialized in the module itself_, the check will pass even if the user's config did not assign that setting.

Example:
`google_calendar` has `required = ('credential_path',)` , but also [`credential_path = None`](blob/3dcc2d7d8f9acd057a1c93ff45b510c60c4772ad/i3pystatus/google_calendar.py#L51)

Now if i3pystatus is launched with just `status.register("google_calendar")`, the module is run and crashes with an unhelpful error message. It should have said `Module 'google_calendar': missing required options: {'credential_path'}`.

This can be remedied by keeping initializations of required settings out of module code, but that's another requirement to be enforced. It's better (?) to make the checks in `SettingsBase` ignore settings that are required and set in the module itself.

enkore:
Good catch.

> If a config in that set is initialized in the module itself, the check will pass even if the user's config did not assign that setting.

It (SettingsBase) should check whether the class has a default value and raise in that case.

PR #427, opened 2016-08-06
Ensure required settings are defined correctly.

facetoe:
Possible fix for #424. If a setting has not been overidden in a subclass
or set in the user's config then raise an exception.


ncoop:
:+1:  Works on the `google_calendar` example.

enkore:
A (regression) test case may be good

facetoe:
This seems to work, however if a setting is defined in a subclass and is set to `None` an exception will still be raised.

PR #436, opened 2016-08-28
Fix issue in network when ipv6 includes subnet suffix

grimpy:
Probably this is caused by new version of netifaces


enkore:
Does #435 also fix this problem?

grimpy:
Yes it does, and looks like a much cleaner solution.

PR #443, opened 2016-08-30
openstack_vms: update novaclient auth

drwahl:
The openstack novaclient has updated it's auth mechanism, deprecating
the method used in this module.  This patch updates the novaclient
authentication call to leverage this new auth method.


enkore:
docs/conf.py causes build failure https://github.com/enkore/i3pystatus/blob/master/docs/conf.py#L34

drwahl:
Doh! Fixed.

Issue #397, opened 2016-06-09
Optimize weather module


terminalmage:
Because ``run()`` is called whenever a callback happens, scrolling the mouse/trackpad in the weather module results in a torrent of unnecessary updates. Updates should instead be moved to a thread, and manual updates assigned to one of the callbacks.

Creating this issue to remind myself to implement this.

terminalmage:
This optimization is now complete for the Weather.com backend, and includes a full rewrite of Weather.com as their XML API is now defunct.

I've begun the re-organization of the Weather Underground backend, but will need to finish this tomorrow as it is past 1am here and way past my bedtime. :smile:

terminalmage:
#451 was opened on Thursday to fulfill this.

PR #446, opened 2016-09-06
Fix traceback in weather.com weather backend

terminalmage:
It seems that this XML feed may have been changed, as it now returns an
error saying that the location is not valid.

This fixes the traceback when we try to run str functions on None when
the XML keys don't exist in the return data. However, the long term fix
will be to either update the backend to use the updated formatting
(which may be hard as this XML feed appears to be undocumented), or to
simply retire this backend.

### DIFF
--- a/i3pystatus/weather/weathercom.py
+++ b/i3pystatus/weather/weathercom.py
@@ -83,18 +83,26 @@ class Weathercom(Backend):
             xml = handler.read().decode(charset)
         doc = ElementTree.XML(xml)
 
+        def _findtext(xmlpath):
+            '''
+            return the contents of the key at the specified path, or an empty
+            string if the key is not present
+            '''
+            value = doc.findtext(xmlpath)
+            return value if value is not None else ''
+
         # Cut off the timezone from the end of the string (it's after the last
         # space, hence the use of rpartition). International timezones (or ones
         # outside the system locale) don't seem to be handled well by
         # datetime.datetime.strptime().
         try:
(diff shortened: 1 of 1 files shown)


terminalmage:
From watching the requests made when you go to weather.com in the browser, they are using a newer API internally. They seem to be generating an API key on their side to use for API requests. I haven't observed it enough to tell whether or not (or how often) the API key is rotated.

I'll probably need to update the weather.com backend to use this new API, but the question then is what to do about the API key. If every i3pystatus user starts using the same API key, it is probably going to be disabled. So, would it be OK to rely on users to do something like this to get a usable API key?

```
curl -s https://weather.com | grep -i apiKey | grep -v -
          'apiKey': 'c1ea9f47f6a88b9acb43aba7faf389d4',
```

enkore:
This could be done in `init()` or, if necessary, periodically in `run()`, instead of manual key generation.

terminalmage:
Yeah, good call, I like that idea much better than asking someone to get it manually. Doing it in the ``init()`` would insulate the code from the API key changing periodically as well. Great idea!

terminalmage:
I'll try to get a pull request for this in the next couple days.

terminalmage:
@enkore I have rewritten the weather.com backend to use the new API. I'm going to give it at least 24 hours to see if changing conditions result in unanticipated breakage, and I also plan to implement the optimizations needed from https://github.com/enkore/i3pystatus/issues/397 to prevent extra updates when the mouse callbacks are invoked. I will open a separate pull request with those changes and close this one.

enkore:
Thanks! :)

PR #449, opened 2016-09-14
Fix net_speed

gdunstone:
speedtest-cli no longer offers a module interface in the main branch: use speedtest-cli/modularize-2 until it is merged in with master.
I cleaned up net_speed and made it use speedtest-cli properly. It now also gives an indication of upload speed by default.

### DIFF
--- a/i3pystatus/net_speed.py
+++ b/i3pystatus/net_speed.py
@@ -1,5 +1,5 @@
 from i3pystatus import IntervalModule
-import speedtest_cli
+import speedtest
 import requests
 import time
 import os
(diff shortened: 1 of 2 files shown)


drwahl:
You'll need to update https://github.com/enkore/i3pystatus/blob/master/docs/conf.py#L35 reference as well.

i3pystatus/net_speed.py
```diff
@@ -12,23 +12,56 @@
 class NetSpeed(IntervalModule):
     """
     Attempts to provide an estimation of internet speeds.
-    Requires: speedtest_cli
+    Requires: speedtest-cli/modularize-2
```
enkore: Where to get?

drwahl: If I'm not mistaken:
https://github.com/sivel/speedtest-cli/tree/modularize-2

gdunstone: @drwahl Thanks, I havent been quite sure how to include this kind of dependency properly since the modularize-2 branch hasn't been merged yet.
And thanks for the conf thing, I was wondering if there was some other place I forgot to check.

gdunstone:
@drwahl Thanks, I havent been quite sure how to include this kind of dependency properly since the modularize-2 branch hasn't been merged yet.
And thanks for the conf thing, I was wondering if there was some other place I forgot to check.

i3pystatus/net_speed.py
enkore: Maybe add make it a link to the GH branch, because normally "requires" means a PyPI package, while this needs to be installed from source -- you could also show how to install it with pip, e.g. `pip install git+https://github.com/sivel/speedtest-cli.git@modularize-2` (or so)

drwahl:
Another good idea might be to raise an issue with the speedtest-cli project asking them to get a new package into pypi.

gdunstone:
I fixed up the docstring, I think its ok now.
I'm not sure how active the development on speedtest-cli is (last commit was just over a year ago).
I'll open an issue to create a pypi module.

gdunstone:
the modularize-2 branch has now been merged. So the documentation provided is now obsolete.

It won't break anything if that branch is used, but I suspect that it will disappear soon.

PR #451, opened 2016-09-15
Optimize weather module, rewrite Weather.com backend

### LINKED ISSUE #397: Optimize weather module

terminalmage:
Weather.com's XML feed is now defunct, this commit includes a rewritten
Weather.com module which uses the same JSON feed used by the website
itself.

The weather updates now happen in a separate thread rather than
happening in the ``run()`` function. Since the ``run()`` function is
executed before any of the mouse events are run, this was causing the
mouse event callbacks to result in a flurry of weather updates, which
caused the Weather Underground API to be overutilized beyond its rate
limit.

Resolves #397.

### DIFF
--- a/i3pystatus/weather/__init__.py
+++ b/i3pystatus/weather/__init__.py
@@ -1,9 +1,52 @@
+import json
+import re
+import threading
+import time
+from urllib.request import urlopen
+
 from i3pystatus import SettingsBase, IntervalModule, formatp
 from i3pystatus.core.util import user_open, internet, require
 
 
-class Backend(SettingsBase):
+class WeatherBackend(SettingsBase):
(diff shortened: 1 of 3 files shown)


terminalmage:
Just added a commit because it looks like Weather Underground just added a decimal point to their current temp observation within the last couple days. I gave it a couple days though and it never showed anything besides a 0 after the decimal point, so I'm just ignoring the decimal point.

This should be good to merge.

PR #452, opened 2016-09-16
Fix imap re-connection

sileht:
Imap re-connection is not working.

Connection exceptions (socket.error, socket.gaierror, IMAP4.abort) can
be raise during any imap operations, not only on connection
establishment.

Also the idle thread call start() on thread multiple times, but this is
not allowed by python. The thread is not daemonize, so the thread code
can be never runned if the python process is busy. And the module is
located in imaplib2.imaplib2.IMAP4 not imaplib2.IMAP4...

This change fixes all of that.

### DIFF
--- a/i3pystatus/mail/imap.py
+++ b/i3pystatus/mail/imap.py
@@ -1,60 +1,18 @@
 try:
-    from imaplib2 import IMAP4, IMAP4_SSL
+    from imaplib2.imaplib2 import IMAP4, IMAP4_SSL
     use_idle = True
-except:
+except ImportError:
     from imaplib import IMAP4, IMAP4_SSL
     use_idle = False
+import contextlib
+import time
 import socket
-from threading import *
+from threading import Thread
 
 from i3pystatus.mail import Backend
 
 
-# This is the threading object that does all the waiting on
-# the event
-class Idler(object):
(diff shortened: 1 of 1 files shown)


i3pystatus/mail/imap.py
```diff
@@ -1,61 +1,18 @@
 try:
-    from imaplib2 import IMAP4, IMAP4_SSL
+    from imaplib2.imaplib2 import IMAP4, IMAP4_SSL
```
enkore: Good catch. Looks like a bug in imaplib2: imaplib2/__init__.py, it uses `from imaplib2 import *`, while it should be `from .imaplib2 import *` or `from imaplib2.imaplib2 import *`.

Can you cahgen the "except:" v- to something more specific (like `except ImportError`)?

i3pystatus/mail/imap.py
```diff
@@ -84,47 +41,56 @@ def init(self):
         if self.ssl:
             self.imap_class = IMAP4_SSL
 
-        self.conn = self.get_connection()
```
enkore: Interesting, seems this was off for quite some time, since all other code uses self.connection, not self.conn

i3pystatus/mail/imap.py
sileht: Done

PR #455, opened 2016-09-19
Add a GitHub Status module

terminalmage:
This adds a new module for displaying the status of github.com via
GitHub's Status API.


facetoe:
I wrote a simple Github module a while ago - https://github.com/enkore/i3pystatus/blob/master/i3pystatus/github.py. Maybe merge your added functionality into that module? No reason to have two Github modules.

terminalmage:
BTW, my reasoning for proposing changing the ``open_github`` event to a double-click is because a single click already updates the module (via calling ``run()``). This is also why I've been writing my modules lately to handle all their updates in a separate thread. I've found that click events can be interfered with (or not even run at all) when ``run()`` makes changes to ``self.output``.

facetoe:
Yup, sounds good to me @terminalmage.

terminalmage:
Cool, I'll make the changes over the weekend. Thanks for your input.

terminalmage:
@facetoe OK, I've merged the functionality from the two modules. Would you like to take a look and perhaps test it for me? I've been using it for about a day now and I think I've got it in good enough shape to merge.

docs/configuration.rst
```diff
@@ -536,33 +536,83 @@ service that is not traditionally required for web browsing:
...
+       # Requires the keyrings.alt package
+       from keyrings.alt.file import PlaintextKeyring
+       status.register('github', keyring_backend=PlaintextKeyring())
 
+.. _KWallet: http://www.kde.org/
+.. _SecretService: https://specifications.freedesktop.org/secret-service/re01.html
+.. _RuntimeError: https://docs.python.org/3/library/exceptions.html#RuntimeError
+.. _keyring: https://pypi.python.org/pypi/keyring
+.. _keyrings.alt: https://pypi.python.org/pypi/keyrings.alt
+.. _keyringrc.cfg: http://pythonhosted.org/keyring/#customize-your-keyring-by-config-file
```
enkore: Great docs!

terminalmage: I write RST at my day job. :smile:

terminalmage:
@facetoe I made one enhancement to the notification support in this PR. The module now tracks the ``id`` value for each notification so that we can detect new notifications (and trigger a desktop notification to alert the user). A simple count of the number of elements in the response wouldn't be sufficient, as someone could mark a few notifications as read and this would affect the notification count.

Another thing I noticed while testing is that the ``/notifications`` API endpoint only seems to return the newest 50 notifications. I noticed this because I haven't been marking my notifications as read for a while and had over 100 of them, but in my testing I only ever showed 50 notifications until I started marking them as read and dipped below 50. According to the [api documentation](https://developer.github.com/v3/activity/notifications/#list-your-notifications) there doesn't appear to be a way to request more than 50, but the 50 number isn't even _in_ the documentation. I'm contacting GitHub for clarification on this, and based on what I hear back, this would be useful information to add to the docstring.

terminalmage:
OK, I had a suspiscion that this was just [pagination](https://developer.github.com/v3/#pagination), but the difference of 50 to what the pagination docs say is should be the limit (30) made me think there was something else going on here. After emailing back and forth with GitHub support, they confirmed that the ``/notifications`` endpoint does indeed default to 50 and not 30, an undocumented departure from the documented pagination behavior (which they'll hopefully document for the future).

According to the [rate-limiting docs](https://developer.github.com/v3/#rate-limiting), authenticated requests are rate-limited at 5000 per hour, which should be more than enough for us to retrieve all notifications via successive ``/noification`` endpoint hits. The rate limiting information, as well as API links to get to the next page of notification results, are in the X-headers:

```zsh
% curl --silent -i 'https://api.github.com/notifications?access_token=<REDACTED>&all=1' | egrep '^(X-RateLimit|Link)'
X-RateLimit-Limit: 5000
X-RateLimit-Remaining: 4987
X-RateLimit-Reset: 1474923479
Link: <https://api.github.com/notifications?access_token=<REDACTED>&all=1&page=2>; rel="next", <https://api.github.com/notifications?access_token=<REDACTED>&all=1&page=3>; rel="last"
```

Should be an easy modification to the module to check the response headers. From what I can see, the ``Link`` header is not present when you are currently on the last page of results, so we'll have an easy way of telling when to break from the loop.

I'll make the necessary modifications over the next day or two and update this pull request, let's hold off on merging until then.

facetoe:
Awesome @terminalmage! I haven't had a chance to test yet but will try to have a look tonight.

i3pystatus/github.py
```diff
@@ -1,83 +1,544 @@
...
+                # No change, so no notification
+                return
+            self.previous_status = response
+
+        if self.previous_status is None:
+            # The only way this would happen is if we invoked the right-click
+            # event before we completed the initial status check.
+            return
+
+        self.notify_status_change()
```
facetoe: This method doesn't exist.

i3pystatus/github.py
```diff
@@ -1,83 +1,544 @@
...
+
+    @staticmethod
+    def notify(message):
+        return DesktopNotification(title='GitHub', body=message).display()
+
+    def skip_notify(message):
```
facetoe: I think this is missing a `self` parameter to the method (to access the logger).

i3pystatus/github.py
```diff
@@ -1,83 +1,544 @@
...
+
+        try:
+            if not self.username and not self.password and not self.access_token:
+                # Auth not configured
+                self.logger.debug(
+                    'No auth configured, notifications will not be checked')
+                return False
+
+            if not HAS_REQUESTS:
+                self.log.error(
```
facetoe: Should be `logger` not `log`.

i3pystatus/github.py
```diff
@@ -1,83 +1,544 @@
...
+    def update_unread(self):
+        # Reset the new_unread attribute to prevent spurious notifications
+        self.new_unread = None
+
+        try:
+            if not self.username and not self.password and not self.access_token:
+                # Auth not configured
+                self.logger.debug(
+                    'No auth configured, notifications will not be checked')
+                return False
```
facetoe: Why return `True` and `False` from this method? As far as I can see the return value is never used.

i3pystatus/github.py
```diff
@@ -1,83 +1,544 @@
...
+    format = '{status}[ {unread}][ {update_error}]'
+    status = _default_status
+    colors = _default_colors
+    refresh_icon = '⟳'
+    update_error = '!'
     username = ''
     password = ''
     access_token = ''
-    format = '{unread}'
+    unread_marker = '•'
```
facetoe: The original marker was bigger. Could we change it back to "●"?

terminalmage: I tried 4 or 5 fonts and the marker was so big that it was changing the vertical alignment of the i3bar. If this is not the case for you, then I can just revert it and manually specify it in the module registration.

i3pystatus/github.py
terminalmage: Yep, used to be named differently and I changed the function name but not the ref to it. Good catch!

i3pystatus/github.py
terminalmage: Yeah, this was copypasta'ed from the static method above it and I didn't add the ref to ``self``. Thanks.

i3pystatus/github.py
terminalmage: This is meant to allow potential future development choices to easily know whether or not the function was successful, e.g.

```python
if not self.update_unread():
    do_something_here
```

i3pystatus/github.py
terminalmage: Thanks!

terminalmage:
@facetoe thanks for the review, I just pushed changes that should address all the concerns you raised.

I'll still need a day or so to get the notification support fleshed out per my earlier comment.

i3pystatus/github.py
facetoe: Ahh, I previously had only the marker with no text so I wouldn't have noticed if it messed up the alignment. It's fine if you leave that one and I can manually specify the larger one in my config.

terminalmage: Sure, thanks!

terminalmage:
I had a crazy week last week, but should be able to wrap this up in the next few days.

facetoe:
All good @terminalmage. A couple of times I got the notification about an error however there was nothing in the logs. I haven't had a chance to look into it very deeply though. Apart from that it's working great for me.

terminalmage:
@facetoe that's odd, I am logging errors at the ``error`` loglevel which should be verbose enough to be logged with the default loglevel. You didn't do anything like manually set the loglevel to ``critical``, did you?

I'll look at the instances where I'm setting that error condition and make sure that I'm not doing anything silly like logging at ``debug``.

terminalmage:
@facetoe I just pushed an update that follows the ``Link`` header to get notifications from multiple pages. I also identified one instance in which I was not logging at the ``error`` level when the ``update_error`` formatter is set, and that is when the API response is blank. I changed the loglevel to ``error`` for this.

facetoe:
Awesome @terminalmage, I've pulled the latest changes and will let you know if I see any more of the error notifications.

terminalmage:
@facetoe Cool. If you want to test getting multiple pages of notifications without letting more than 50 unread notifications pile up, you can add ``&all=1`` to the ``ACCESS_TOKEN_AUTH_URL`` global. It doesn't seem to grab *every* notification I've ever had, but it was sufficient enough for me to get over that 50 notification limit to trigger the pagination behavior.

terminalmage:
@facetoe have you had a chance to test with the updated code? I've been using it for a couple weeks now without issue.

facetoe:
Yup I haven't had any issues either. One thing I would suggest though is removing the notification when the module errors out trying to talk to github. I have a pretty unreliable internet connection so it informs me of that quite often! It's probably enough to just log it anyway.

terminalmage:
@facetoe Good point. We're already logging the error, as well as setting the ``{update_error}`` formatter, so this is overkill. I pushed a commit to get rid of that notification. I'll go ahead and merge this. Thanks for the input and for helping to test!

facetoe:
No worries @terminalmage !

PR #461, opened 2016-10-05
pulseaudio: allow to configure the sink

sileht:
Hi,

When we have multiple sinks, this change allows to have all/some of them
available in i3pystatus and to identify which one is currently the
default. Also it allows to configure if we want that the hook
change_sink move all inputs to the new sink or not.

Cheers,


enkore:
Can you rebase this / merge with master?

sileht:
Done

enkore:
Thanks!

PR #463, opened 2016-10-08
Timewarrior module

JindrichPilar:
Module for heads up information for taskwarrior.org/docs/timewarrior/

### DIFF
--- a/i3pystatus/timewarrior.py
+++ b/i3pystatus/timewarrior.py
@@ -41,8 +41,9 @@ class TImewarrior(IntervalModule):
             tracks = loads(tracks_json.decode("utf-8"))
             self.track = tracks[-1]
 
-        except ValueError:
+        except ValueError as error:
             print('Decoding JSON has failed')
+            raise error
 
     def stop_or_continue(self):
         self.loadTrack()
--- a/i3pystatus/timewarrior.py
+++ b/i3pystatus/timewarrior.py
@@ -6,7 +6,7 @@ from datetime import datetime, timezone
 import subprocess
 
 
-class TImewarrior(IntervalModule):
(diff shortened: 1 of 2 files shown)


i3pystatus/timewarrior.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    def loadTrack(self):
+        try:
+            tracks_json = subprocess.check_output(['timew', 'export'])
+            tracks = loads(tracks_json.decode("utf-8"))
+            self.track = tracks[-1]
+
+        except ValueError:
+            print('Decoding JSON has failed')
```
facetoe: Should rethrow an exception here, that way the error will be visible in the status bar and not break things.

JindrichPilar: Error is now re-raised.

i3pystatus/timewarrior.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+    def loadTrack(self):
+        try:
+            tracks_json = subprocess.check_output(['timew', 'export'])
+            tracks = loads(tracks_json.decode("utf-8"))
+            self.track = tracks[-1]
+
+        except ValueError as error:
+            print('Decoding JSON has failed')
```
enkore: log instead of print

JindrichPilar: Print changed to self.logger.exception.

enkore:
Try adding dateutil.parser and dateutil.relativedata to docs/conf.py

See http://i3pystatus.readthedocs.io/en/latest/module.html#handling-dependencies

JindrichPilar:
I have updated ``docs/conf.py``.

PR #471, opened 2016-10-17
online module: Don't fail when offline

hobarrera:
If the system is offline, `socket.gaierror` is raised, the module crashes, and shows `offline` indefinitely.

Catch that exception, and return False.

### DIFF
--- a/i3pystatus/core/util.py
+++ b/i3pystatus/core/util.py
@@ -373,7 +373,7 @@ class internet:
         try:
             socket.create_connection(cls.address, 1).close()
             return True
-        except OSError:
+        except (OSError, socket.gaierror):
             return False
 
 


i3pystatus/core/util.py
```diff
@@ -373,7 +373,7 @@ def __new__(cls):
         try:
             socket.create_connection(cls.address, 1).close()
             return True
-        except OSError:
+        except (OSError, socket.gaierror):
```
enkore: Since Python 3.3 gaierror is also a subclass of OSError, so this should only affect 3.2 and earlier.

hobarrera: I'm using python3.5 and the original code isn't catching gaierror. Is it possible that this was later reverted?

hobarrera:
Any ETA on a next release? I've been waiting for this to reach the stable releases for quite a while 😄

PR #475, opened 2016-10-19
network: add next_if_down option

m45t3r:
This option switch to the next interface (using `cycle_interface()` method) if the current one is down.

The use case is to have a more automatic behavior. In my case, I switch between Ethernet and Wi-Fi constantly because of my work, so switching manually tires me.

The behavior with this option on is a little problematic in the case where both interfaces are down (if keeps switching between them). And the default scroll up/scroll down to switch interfaces does not really work unless all interfaces are up. This is why I set this option to default to `False`.

However, if there is some suggestion on how to improve the behavior I am all ears.

### DIFF
--- a/i3pystatus/network.py
+++ b/i3pystatus/network.py
@@ -286,6 +286,7 @@ class Network(IntervalModule, ColorRangeModule):
         ("round_size", "defines number of digits in round"),
         ("detached_down", "If the interface doesn't exist, display it as if it were down"),
         ("unknown_up", "If the interface is in unknown state, display it as if it were up"),
+        ("next_if_down", "Change to next interface if current one is down"),
     )
 
     interval = 1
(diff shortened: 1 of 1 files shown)


facetoe:
> The behavior with this option on is a little problematic in the case where both interfaces are down 

Maybe you could check this before cycling the interface. If all valid interfaces are down then just do nothing.

m45t3r:
@facetoe this would need a refactoring of some code in Network module.

Basically, instead of defining one interface and cycling between all valid interfaces if this interface does not exist or is down, I would change this parameter to accept a list, so only the interfaces in this list would be cycled. Of course, a new sane default like passing None would detect all interfaces.

m45t3r:
Anyway, even with the current problems this option works for my use case, this is why I am not focusing too much in this issue. Any comments @enkore?

enkore:
What does "it keeps switching between them" mean? As long as it's only with the interval I see no big problem there, it just shouldn't be a busyloop eating 100% CPU ;)

m45t3r:
@enkore, yeah, it is just this, it changes at each update interval. You can test this by increasing interval, it will takes more time to switch.

Issue #442, opened 2016-08-29
"No such file or directory: '/sys/class/backlight/acpi_video0/max_brightness" kills the whole i3pystatus output


teto:
```
Traceback (most recent call last):
  File "/home/teto/.local/bin/i3pystatus", line 9, in <module>
    load_entry_point('i3pystatus', 'console_scripts', 'i3pystatus')()
  File "/home/teto/i3pystatus/i3pystatus/__init__.py", line 47, in main
    imp.load_source(module_name, args.config)
  File "/usr/lib/python3.5/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/home/teto/.i3/myStatus.py", line 51, in <module>
    status.register("backlight",   format="{percentage}%",)
  File "/home/teto/i3pystatus/i3pystatus/core/__init__.py", line 109, in register
    return self.modules.append(module, *args, **kwargs)
  File "/home/teto/i3pystatus/i3pystatus/core/util.py", line 73, in append
    module, *args, **kwargs)
  File "/home/teto/i3pystatus/i3pystatus/core/imputil.py", line 47, in instanciate_class_from_module
    return self.instanciate_class_from_module(self.get_module(module), *args, **kwargs)
  File "/home/teto/i3pystatus/i3pystatus/core/imputil.py", line 45, in instanciate_class_from_module
    return self.get_class(module)(*args, **kwargs)
  File "/home/teto/i3pystatus/i3pystatus/core/modules.py", line 64, in __init__
    super(Module, self).__init__(*args, **kwargs)
  File "/home/teto/i3pystatus/i3pystatus/core/settings.py", line 109, in __init__
    self.init()
  File "/home/teto/i3pystatus/i3pystatus/backlight.py", line 50, in init
    with open(self.base_path + maxbfile, "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/sys/class/backlight/acpi_video0/max_brightness'
```

At first I had not installed xbacklight and I had a "module error", then I installed xbacklight and got an the uncaught past error (from ".xsession-errors") which kills the whole i3pystatus "status_command process exited unexpectedly". It would be nice to catch it so that it doesn't kill the whole status.

[...]

PR #483, opened 2016-10-26
Backlight: fix crash in #442

hcpl:
Monkey patches the `run()` method to use `format_no_backlight` when /sys/class/backlight/ is empty.

### DIFF
--- a/i3pystatus/backlight.py
+++ b/i3pystatus/backlight.py
@@ -48,7 +48,7 @@ class Backlight(File):
         backlight_entries = sorted(glob.glob(self.base_path))
 
         if len(backlight_entries) == 0:
-            self.run = run_no_backlight
+            self.run = self.run_no_backlight
             super().init()
             return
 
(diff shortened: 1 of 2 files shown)


hcpl:
The only issue now is defining reasonable values for `cdict`:
```python
cdict = {
    "brightness": -1,
    "max_brightness": -1,
    "percentage": -1
}
```
@enkore, @teto, any thoughts about this?

m45t3r:
I think this was fixed in #481, I am wrong?

hcpl:
@m45t3r, glob expansion returns an empty list if directory is empty, so indexing crashes.
> @teto:
> I pulled the changes but it still kills the whole output as
> "/sys/class/backlight" is empty.

m45t3r:
Hm... I only tested the garbage case, not the empty case. Makes sense.

hcpl:
@enkore, I suggest that this module should be tested to prevent future bugs.

hcpl:
@teto [reported](https://github.com/enkore/i3pystatus/issues/442#issuecomment-257096718) that everything is good. I think this is ready to merge now.

enkore:
Very nice, thanks!

Issue #482, opened 2016-10-25
Sending SIGUSR1 when redshift module is active disables it instead of refreshing i3pystatus


m45t3r:
Both i3pystatus and redshift use SIGUSR1 as a special signal. When you try to send SIGUSR1 to i3pystatus process, redshift captures the signal, activating/deactivating inhibit, since the thread inherits the signal handler.

I don't know a good way to fix this for now. We could use `multiprocessing` instead of `threading`, since in this case the process handler is maintained with parent instead of child, however in this case the module will have to be rewritten to allow communication between the two process. Another way would be to rewrite the module to implement the process parsing ourselves, removing the GLib dependency and allowing us to choose which signal to ignore.

Any suggestion are welcome.

richese:
This call added in 3.3 might be usefull: [signal.pthread_sigmask](https://docs.python.org/3/library/signal.html#signal.pthread_sigmask). It can block a thread from receiving certain signals.

Something like this could work if [redshift:73](https://github.com/enkore/i3pystatus/blob/115bd4d34f76bd991ad2ec7a150794a8b8da2e12/i3pystatus/redshift.py#L73) is replaced  with a custom callback.
```python
def thread_callback(self):
    signal.pthread_sigmask(signal.SIG_BLOCK, [signal.SIGUSR1])
    Gtk.main()
```

m45t3r:
@richese I will try this one, thanks.

BTW, any notice about issue #430?

m45t3r:
@richese Does not work. I tried to block all signals with `range(1, signal.NSIG)`, or using `signal.SIG_SETMASK` too, without success.

richese:
Well, it was worth a try.
The USR1 signal handler is added in [constructor]( https://github.com/jonls/redshift/blob/639432fcfa3884c1e2169911c581120b7a7981f8/src/redshift-gtk/statusicon.py#L115) of RedshiftController and I have no idea how to remove or mask it in GLib without the return value from [GLib.unix_signal_add()](http://lazka.github.io/pgi-docs/#GLib-2.0/functions.html#GLib.unix_signal_add).

m45t3r:
@richese I will probably rewrite this module, now I understand what ReshiftController does so it is doable. I think it is possible to implement using pure Python without any additional dependencies except redshift binary itself.

m45t3r:
Should be fixed in #484.

PR #484, opened 2016-10-26
Refactor redshift module

m45t3r:
I am testing this for now, however it seems quite good. No more dependency in redshft-gtk, and it as far I tested it fixes the issue #482.

### DIFF
--- a/i3pystatus/redshift.py
+++ b/i3pystatus/redshift.py
@@ -32,7 +32,14 @@ class RedshiftController(threading.Thread):
 
         env = os.environ.copy()
         env['LANG'] = env['LANGUAGE'] = env['LC_ALL'] = env['LC_MESSAGES'] = 'C'
-        self._p = Popen(cmd, env=env, stdout=PIPE, bufsize=1, universal_newlines=True)
+
+        self._params = {
+            "args": cmd,
+            "env": env,
+            "bufsize": 1,
+            "stdout": PIPE,
+            "universal_newlines": True,
+        }
 
     def parse_output(self, line):
         """Convert output to key value pairs"""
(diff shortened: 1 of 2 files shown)


i3pystatus/redshift.py
```diff
@@ -1,14 +1,110 @@
...
+        # Regex to parse output
+        self._regex = re.compile(r'([\w ]+): (.+)')
+
+        # Initialize state variables
+        self._inhibited = False
+        self._temperature = 0
+        self._period = 'Unknown'
+        self._location = (0.0, 0.0)
+
+        cmd = ["redshift"]
```
richese: I think you forgot to add user provided `args` to `cmd`.

m45t3r:
@richese I think this is mostly fault of redshift itself:

```
$ redshift -v | sleep 1 | killall redshift && echo $?
[1]    32106 terminated  redshift -v | 
       32107 done        sleep 1 | 
       32108 done        killall redshift
0
```
As you can see, it returns 0 even if it is killed by force. I can force an error if the process exited, though.

m45t3r:
I just report an error if redshift exit for any motive, since to get the return code of the process would complicate the code a lot (I am running redshift in a separate thread, so I would need some kind of communication). And I think this is fine, since redshift shouldn't exit anyway during execution.

What do you think @richese?

richese:
@m45t3r 
Yeah, I think that's fine. Now it gives some feedback if you try to start redshift with invalid arguments or it exits for another reason.

m45t3r:
I am using this for 2 Days already and didn't found any other bug. I think it is ready to merge.

@enkore, what do you think?

teto:
Could you inhibit as a default please if it's not done already ? I don't know what others think but I find it disruptive cf https://github.com/enkore/i3pystatus/issues/489

m45t3r:
@enkore, any comments?

enkore:
Code-wise this looks good but needs a merge of the origin/master branch or a rebase on top of it.

m45t3r:
@enkore Rebased.

Issue #490, opened 2016-10-30
Error: status_command process exited unexpectedly


jcf:
When something goes wrong with a single module the entire status line can be brought down.

In my case the "Master" mixer can't be found, which in turn blows up alsaaudio, which then takes down all of my modules.

```
Traceback (most recent call last):
  File "/home/jcf/.config/i3/status.py", line 18, in <module>
    format=smaller("Vol {volume} / 100"))
  File "/usr/lib/python3.5/site-packages/i3pystatus/core/__init__.py", line 110, in register
    return self.modules.append(module, *args, **kwargs)
  File "/usr/lib/python3.5/site-packages/i3pystatus/core/util.py", line 73, in append
    module, *args, **kwargs)
  File "/usr/lib/python3.5/site-packages/i3pystatus/core/imputil.py", line 47, in instanciate_class_from_module
    return self.instanciate_class_from_module(self.get_module(module), *args, **kwargs)
  File "/usr/lib/python3.5/site-packages/i3pystatus/core/imputil.py", line 45, in instanciate_class_from_module
    return self.get_class(module)(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/i3pystatus/core/modules.py", line 64, in __init__
    super(Module, self).__init__(*args, **kwargs)
  File "/usr/lib/python3.5/site-packages/i3pystatus/core/settings.py", line 109, in __init__
    self.init()
  File "/usr/lib/python3.5/site-packages/i3pystatus/alsa.py", line 59, in init
    self.create_mixer()
  File "/usr/lib/python3.5/site-packages/i3pystatus/alsa.py", line 77, in create_mixer
    control=self.mixer, id=self.mixer_id, cardindex=self.card)
alsaaudio.ALSAAudioError: Unable to find mixer control Master,0 [hw:0]
```

I'll see if I can fix the underlying issue with my audio environment, but wonder if the experience could be a little more polished when something like this occurs?
I'm not familiar with the internals of i3pystatus so can only guess if it's even possible.

Thanks for i3pystatus!

jcf:
Thanks, @facetoe!

PR #491, opened 2016-11-01
Catch exceptions thrown in module.init().

facetoe:
Possible solution for #490

### DIFF
--- a/i3pystatus/core/__init__.py
+++ b/i3pystatus/core/__init__.py
@@ -9,6 +9,7 @@ from i3pystatus.core.imputil import ClassFinder
 from i3pystatus.core.modules import Module
 
 DEFAULT_LOG_FORMAT = '%(asctime)s [%(levelname)-8s][%(name)s %(lineno)d] %(message)s'
+log = logging.getLogger(__name__)
 
 
 class CommandEndpoint:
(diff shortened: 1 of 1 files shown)


i3pystatus/core/__init__.py
```diff
@@ -106,22 +106,32 @@ def register(self, module, *args, **kwargs):
...
             if import_error.name and not import_error.path and isinstance(module, str):
                 # This is a package/module not found exception raised by importing a module on-the-fly
                 return self.modules.append(Text(
-                    color="#FF0000",
+                    color=err_colour,
                     text="{i3py_mod}: Missing Python module '{missing_module}'".format(
                         i3py_mod=module,
                         missing_module=import_error.name)))
             else:
                 raise import_error
```
hcpl: The try block doesn't catch this ImportError

facetoe: True, although I'm not sure why the import error is reraised in the first place. It does seem nicer to catch all the exceptions here and alert the user via the status bar.

hcpl: Probably, this must be a very unusual case, so the author reraised it as a TODO reminder, but eventually forgot it :)

enkore: I wrote it to only handle import errors that happened if you specified a module as a string. if you specify an instance you'd already have imported it and it would already have crashed before i3pystatus got involved.

Not sure why I wrote it to *only* handle this kind of error and not all ImportErrors.

facetoe: OK cool, I'll modify the try block to catch all possible errors tonight.

i3pystatus/core/__init__.py
```diff
@@ -106,22 +106,32 @@ def register(self, module, *args, **kwargs):
...
                 text=configuration_error.message))
+        except Exception as e:
+            return self.modules.append(Text(
+                color=err_colour,
+                text="{i3py_mod}: Fatal Error - {ex}({msg})".format(
+                    i3py_mod=module,
+                    ex=e.__class__.__name__,
+                    msg=e
+                )
+            ))
```
hcpl: How about `Status` having an `abort_on_error` option (default `False`) and used like
```python
if self.abort_on_error:
    raise e
else:
    return self.modules.append(Text(...))
```
This would be useful for debugging broken modules.

hcpl: How about `Status` having an `abort_on_error` option (default `False`) and used like
```python
if self.abort_on_error:
    raise e
else:
    return self.modules.append(Text(...))
```
This would be useful for debugging broken modules.

facetoe: Another option could be to just log the stacktrace.

enkore: Logging the exception makes sense, yeah (more detail than the in-bar text).

abort_on_error doesn't sound very helpful to me.

hcpl: Agreed, logging works for multiple modules at once.

facetoe: Ok cool, I've pushed a new commit that catches all exceptions, logs them and then displays the error in the status bar.

teto:
Really great thanks !

PR #492, opened 2016-11-03
Xkblayout: refine formatting and switching layouts

hcpl:
Introduces PyPI package xkbgroup as a dependency.

Changes include:

* Add more formatters.
  - Backwards-incompatible: ``{name}`` now refers to the layout group
    name, such as 'English (US)' or 'German'.
  - To show the old ``{name}`` use ``{symbols::,}`` (this is the
    advanced formatting syntax employed by xkbgroup package)
    + For details on formatting see
      https://github.com/hcpl/xkbgroup/blob/master/docs/formatting.rst.
* Now Xkblayout can switch groups and you can toggle groups with
  keyboard at the same time.

### DIFF
--- a/i3pystatus/xkblayout.py
+++ b/i3pystatus/xkblayout.py
@@ -16,6 +16,8 @@ class Xkblayout(IntervalModule):
     ``layouts`` can be stated with or without variants,
     e.g.: ``status.register("xkblayout", layouts=["de neo", "de"])``
 
+    Requires xkbgroup (from PyPI)
+
     .. rubric:: Available formatters
 
     * `{num}` — current group number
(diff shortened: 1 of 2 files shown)


i3pystatus/xkblayout.py
```diff
@@ -1,5 +1,8 @@
-from i3pystatus import IntervalModule
 import subprocess
+from itertools import zip_longest
+
+from i3pystatus import IntervalModule
+from xkbgroup import XKeyboard
```
enkore: Can you document this in the module doc string?

hcpl: Done.

PR #499, opened 2016-11-30
Rewrite DNF updates module to use DNF's Python API

### LINKED ISSUE #498: DNF Updates module always displays "Updates: 0"
Running Fedora 25
all other modules seem to be working fine.
DNF Update always displays "Updates: 0"
Here is the updates section from my config file:

status.register("updates",
                        format = "Updates: {count}",
                        color = "#689d6a",
                        color = "#00DD00",,
                        format_no_updates = "No updates",
                        color_no_updates = "#ebdbb2",
                        format_working = None,
                        color_working = None,
                        backends = [dnf.Dnf()],)

Thanks!

terminalmage:
This is a lot easier than trying to parse the DNF CLI output.

Fixes #498.

### DIFF
--- a/i3pystatus/updates/__init__.py
+++ b/i3pystatus/updates/__init__.py
@@ -126,7 +126,10 @@ class Updates(Module):
         for backend in self.backends:
             name = backend.__class__.__name__
             updates, notif_body = backend.updates
-            updates_count += updates
+            try:
+                updates_count += updates
+            except TypeError:
+                pass
             self.data[name] = updates
             self.notif_body[name] = notif_body or ""
 
--- a/i3pystatus/updates/dnf.py
+++ b/i3pystatus/updates/dnf.py
@@ -1,9 +1,17 @@
 from i3pystatus.updates import Backend
+import sys
+
+# Remove first dir from sys.path to avoid shadowing dnf module from
+# site-packages dir when this module executed directly on the CLI.
+__module_dir = sys.path.pop(0)
 try:
     import dnf
     HAS_DNF_BINDINGS = True
 except ImportError:
     HAS_DNF_BINDINGS = False
+finally:
+    # Replace the directory we popped earlier
+    sys.path.insert(0, __module_dir)
 
 
 class Dnf(Backend):
--- a/i3pystatus/updates/dnf.py
+++ b/i3pystatus/updates/dnf.py
@@ -1,33 +1,40 @@
-from i3pystatus.core.command import run_through_shell
 from i3pystatus.updates import Backend
-from re import split, sub
+try:
+    import dnf
+    HAS_DNF_BINDINGS = True
+except ImportError:
+    HAS_DNF_BINDINGS = False
 
 
 class Dnf(Backend):
     """
-    Gets updates for RPM-based distributions with `dnf check-update`.
+    Gets updates for RPM-based distributions using the `DNF API`_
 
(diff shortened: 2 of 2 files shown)


terminalmage:
@enkore We should probably rename the update checker module to something else, since it is named the same as the DNF's built-in python bindings and this causes problems when you try to run the module on the CLI (e.g. ``python3 /path/to/i3pystatus/updates/dnf.py``).

enkore:
`python -m i3pystatus.updates.dnf` should work, if it's installed / installed in dev mode (pip -e / setup.py develop).

terminalmage:
The module works just fine. It's what happens when you run dnf.py directly on the CLI that is the problem:

```
[root@f25 i3]# python3 ~/i3pystatus/i3pystatus/updates/dnf.py
Updates: ?

module 'dnf' has no attribute 'Base'
```

This is because the parent directory of the module being run is first in ``sys.path``, and thus ``import dnf`` ends up importing ``i3pystatus.updates.dnf`` instead of the ``dnf`` module from ``site-packages``.

terminalmage:
@enkore I added another commit which pops off the first element of ``sys.path`` just long enough to ``import dnf`` before replacing it. This allows for the ``dnf`` module to be imported from ``site-packages`` rather than from ``i3pystatus.updates``.

It's a bit of a hack, but it keeps us from needing to do something like rename the module to avoid shadowing. What do you think?

enkore:
Neat hack indeed. I don't have a name that wouldn't be awkward / inconsistent with the others, let's keep it

ncoop:
👍 

But if you are running i3pystatus from a pyenv virtual environment, you may find it unable to import `dnf`. You need to have a pyenv that inherits the system packages and uses the system binary.

```shell
pyenv_name=venv-system   # name this whatever
pyenv virtualenv --system-site-packages --python=/usr/bin/python3 $pyenv_name
```

You know it worked if this throws no error:
```
PYENV_VERSION=$pyenv_name python -c 'import dnf'
```

Now you have to reinstall i3pystatus and the dependencies of the modules you use:
```shell
PYENV_VERSION=$pyenv_name $i3pystatus_repo/setup.py build install
PYENV_VERSION=$pyenv_name pip install colour netifaces psutil
```

And use the right pyenv when you invoke i3pystatus:
```
bar {
  position top
  status_command PYENV_VERSION=$pyenv_name python $path_to_your_config
}
```
None of this should require root.

This should only be necessary if you run i3pystatus from a pyenv and require a module (like dnf) which is only available from the system install. (If Red Hat likes Python so much, they should put their stuff in pypi modules like everyone else.)

pyenv is no concern of i3pystatus, but this should probably be documented somehow.

terminalmage:
It would still happen in a non-pyenv virtualenv, if ``--no-site-packages`` is used:

```
[root@f25 ~]# cat /etc/redhat-release
Fedora release 25 (Twenty Five)
[root@f25 ~]# virtualenv --no-site-packages test
Using base prefix '/usr'
New python executable in /root/test/bin/python3
Also creating executable in /root/test/bin/python
Installing setuptools, pip, wheel...done.
[root@f25 ~]# test/bin/python -c 'import dnf'
  File "<string>", line 1, in <module>
ImportError: No module named 'dnf'
```

I just opened https://github.com/enkore/i3pystatus/pull/502 to document this. Thanks for bringing it up.

PR #500, opened 2016-12-01
gpu_temp: setting "display_if" default value fixed

PolarNick239:
gpu_temp setting "display_if" (introduced in https://github.com/enkore/i3pystatus/pull/382) default value expected to be string (like in https://github.com/enkore/i3pystatus/pull/383)

Error message:
```
GPUTemperature: TypeError: eval() arg 1 must be a string, bytes or code object
```


enkore:
Good catch

PR #503, opened 2016-12-07
Add configuration for graph direction

nkoester:
Adds a new setting fort the cpu_usage_graph which allows to change the direction of the graph from the default 'left to right to 'right to left'


i3pystatus/cpu_usage_graph.py
```diff
@@ -27,12 +27,14 @@ class CpuUsageGraph(CpuUsage, ColorRangeModule):
...
         ("graph_width", "Width of the cpu usage graph"),
         ("graph_style", "Graph style ('blocks', 'braille-fill', 'braille-peak', or 'braille-snake')"),
+        ("direction" , "Graph running direction ('lr', 'rl')"),
     )
 
     graph_width = 15
     graph_style = 'blocks'
     format = '{cpu_graph}'
     cpu = 'usage_cpu'
+    direction = 'lr'
```
facetoe: Maybe use full words here, "lr" and "rl" is  a bit cryptic.

facetoe: Should also verify that the value passed for `direction` is valid.

Issue #505, opened 2016-12-20
Bug in yaourt update backend


jotasi:
When looking into the issue I described in #504, it seemed to me that actually the case of `self.aur_only=True` is actually completely broken. For that case, the output of the shell command (which is a `string`) is parsed using:

    if(self.aur_only):
        out = [line for line in out if line.startswith("aur")]

But as far as I can see, this will always result in an empty `out`, as this will iterate out character by character rather than line by line and thus the condition `line.startswith("aur")` will never be met. I'll submit a pull request fixing this and #504 today.

enkore:
Fixed by #506

Issue #504, opened 2016-12-18
Inconsistent return value types for update backends


jotasi:
When trying to add a new backend to check for updates, I noticed that the return value types are inconsistent across different update backends, leading to breakage of the DesktopNotification in the report method for some of the backends.

While all backends return a number of updates, the listing of the available updates is one string containing one update per line for the backends `cower`, `dnf`, and `pacman`. `aptget` on the other hand returns a list of one line strings. `yaourt` seems to be even inconsistent with itself, returning a string, when `aur_only=False` and a list otherwise. The problem with this is, that `"\n".join(self.notif_body.values())` in line 156 of `i3pystatus/updates/__init__.py` fails for backends that return a list, as `self.notif_body.values()` is then a list of list rather than a list of strings.

I would be happy to submit a pull request cleaning this up, but I first wanted to clarify, which return value is considered correct. As most backends return a string, I'd guess, that this is the way to go. Is this correct?

enkore:
Fixed by#507

PR #506, opened 2016-12-20
Made updates method return types consistent and fixed bug in yaourt backend

jotasi:
Changed the return types of aptget and yaourt to (int, string) from (int, list) to ensure consistency across all update backends and fixing the desktop notification for available updates, fixing #504.
Also fixed the bug in the yaourt backend that broke the `self.aur_only==True` case (see #505),

### DIFF
--- a/i3pystatus/updates/aptget.py
+++ b/i3pystatus/updates/aptget.py
@@ -23,9 +23,9 @@ class AptGet(Backend):
         command = "apt-get upgrade -s -o Dir::State::Lists=" + cache_dir
         apt = run_through_shell(command.split())
 
-        out = apt.out.splitlines()
-        out = [line[5:] for line in out if line.startswith("Inst ")]
-        return len(out), out
+        out = apt.out.splitlines(True)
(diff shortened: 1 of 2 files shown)


ncoop:
This does unbreak yaourt. 👍 

Incidentally there is some old cruft in the pydoc:
```
    If you want to count both pacman and aur packages with this module you can
    set the variable count_only_aur = False like this:
```
`count_only_aur` isn't a variable, and the user doesn't care anyway if it's passed unnamed.

jotasi:
I can change that sentence as well. How about:

    To count both pacman and aur packages, pass False for the parameter count_only_aur in the constructor:

ncoop:
Yes, but I mean something simpler like
```
To count both pacman and aur packages, pass False in the constructor:
```

jotasi:
@ncoop Ok. I changed the documentation to the phrase you suggested.

hasB4K:
I just noticed that bug in the yaourt module! Thank you @jotasi :). I think this pull request should be accepted.

enkore:
Thanks for fixing this!

PR #507, opened 2016-12-20
Added packagekit backend for update notifications

jotasi:
This adds packagekit to the available update backends. As packagekit is used or at least can be used in many different Linux distributions (Standard in e.g. Fedora and OpenSUSE), this could be handy.


enkore:
Neat!

PR #510, opened 2016-12-25
Allow empty stdout on shell module

fmartingr:
I have some `shell` scripts that does not have output for some cases, and the default output for the `shell` module is kinda useless for end results in my opinion.

I've modified the shell module with a new argument to allow having empty output supressing the "Command X returned status code Y" text, defaults to `False` to don't mess up actual configurations for people.

### DIFF
--- a/i3pystatus/shell.py
+++ b/i3pystatus/shell.py
@@ -1,6 +1,5 @@
 from i3pystatus import IntervalModule
 from i3pystatus.core.command import run_through_shell
-import logging
 
 
 class Shell(IntervalModule):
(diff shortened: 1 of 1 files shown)


i3pystatus/shell.py
```diff
@@ -14,9 +13,11 @@ class Shell(IntervalModule):
 
     color = "#FFFFFF"
     error_color = "#FF0000"
+    ignore_emtpy_stdout = False
```
hcpl: Typo?

fmartingr: Damn symbol auto complete. Thanks for pointing this out @hcpl, making another PR now :)

hcpl: You're welcome.

Issue #474, opened 2016-10-19
Spotify module should support other players


m45t3r:
For some reason, Spotify module is artificially limited to only work with Spotify, even if it uses [playerctl](https://github.com/acrisci/playerctl) that support many more players (from README: "spotify, vlc, audacious, bmp, xmms2, and others"; actually it is based on MPRIS protocol that is used by various other players).

Just by not passing "spotify" in [#L78](https://github.com/enkore/i3pystatus/blob/master/i3pystatus/spotify.py#L78) it would allow it to auto detect any compatible running player, allowing the user to set this variable would allow anyone to control their favorite player. I don't know why this modules is design like this, however by just one small modification it would make it more flexible. I had to extend this module to use it with [gpmdp](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/issues), like you can [see here in my dotfiles](https://github.com/m45t3r/dotfiles/blob/master/i3/.config/i3pystatus/config.py#L23).

The module has some other issues too, like not setting gi.required_version("Playerctl", "1.0") (so you get a warning from gi) and catching a generic Exception. So I propose a small refactoring in this module. I could push a PR in this weekend if someone concurs with me.

P.S.: there is now_playing module too, however at least with gpmdp, playerctl seems to work better than dbus interface.

m45t3r:
Fixed in #476.

PR #514, opened 2016-12-29
Add DBus callbacks to now_playing

ncoop:
Similar to #513, you can manipulate the MPRIS Player interface with the `player_command` and `player_prop` callbacks.
https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html

And like 513, this makes the `playpause` and `next_song` callbacks redundant. They can easily be removed.

Unlike socket connections, experimenting with dbus from the shell is not so straightforward, but you can do a lot with `busctl`.

Also, I propose renaming this module to something like `mpris`.

### DIFF
--- a/i3pystatus/now_playing.py
+++ b/i3pystatus/now_playing.py
@@ -33,10 +33,13 @@ track.
     .. rubric:: Available formatters (uses :ref:`formatp`)
 
     * `{title}` — (the title of the current song)
-    * `{album}` — (the album of the current song, can be an empty string (e.g. for online streams))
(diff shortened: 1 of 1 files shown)


ncoop:
Was just looking at issues/prs and saw from #474/#476 that `now_playing` and `playerctl` are very similar. One uses DBus and the other GI? @m45t3r?

m45t3r:
Yeah, they're quite similar. I think one of them could be removed, however this depends on @enkore.

enkore:
I dunno about the difference, but they do look quite similar.

PR #521, opened 2017-01-11
external_ip: add external ip module

hasB4K:
Hello,

This module allow an user to display his current external ip along with the country associated with it.
This module requires the PyPI package `GeoIP`.

Regards,

##
Mathis 'hasB4K' Felardos

### DIFF
--- a/i3pystatus/external_ip.py
+++ b/i3pystatus/external_ip.py
@@ -1,4 +1,5 @@
 from i3pystatus import IntervalModule, formatp
+from i3pystatus.core.util import internet, require
 
 import GeoIP
 import urllib.request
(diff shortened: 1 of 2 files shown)


i3pystatus/external_ip.py
```diff
@@ -0,0 +1,80 @@
...
+    ip_website = "https://api.ipify.org"
+    timeout = 5
+    color = "#FFFFFF"
+    color_hide = "#FFFF00"
+    color_down = "#FF0000"
+
+    on_leftclick = "switch_hide"
+    on_rightclick = "run"
+
+    def run(self):
```
facetoe: Should add the `@require(internet)` decorator so this method is not run when there is no internet connectivity.

hasB4K:
Hi @facetoe, 

This is fixed :). Howewer I have noticed a 'bug' in `require(internet)`. I have explain the issue in my PR #522.

Regards,

Issue #234, opened 2015-08-10
@require(internet) won't work behind a corporate proxy


paulollivier:
Hi,

After wondering why the `i3pystatus.updates` module wasn't reporting any available update when there is, I looked at the code, and saw `i3pystatus/core/util.py:362` where you are making a connection to a google dns server.

The problem is: outside DNS is blocked in many corporate networks.

I was wondering why you choose this way to check for an internet connection, and if you were open to the idea of merging a PR using HTTP, for instance.
Or maybe be able to specify the `host:port` couple to be used when checking for internet access, since the protocol used doesn't seem to be of interest in the current implementation?

Thank you for your time.

enkore:
I did that because it's a very fast and very low traffic method. But yeah I see how it can cause trouble with some tight firewall setups. I'd say changing the target host just to "google.com" and the port to 80 should work well with almost any firewall.

paulollivier:
Yeah, I changed the method to try to connect to one of our internal servers. I was just wondering if it was possible to implement it as a configuration option.

I gather from your response that you are not fond of getting a merge request related to this?

facetoe:
It might be tricky to implement as it's a function without arguments passed to the `require` decorator. Any ideas on how the configuration could be implemented?

t3zla:
Would it not work to just change it to 8.8.8.8?

paulollivier:
Woopsie, sorry, I missed the reply notification.

@facetoe: To be honest, the only way I have found (yet) implies the use of global variables to save the configuration. I fear that a major refactoring would be needed to make it "The Right Way"™.

richese:
How about changing `internet` to a class and making the adress a class variable?
This variable can be set from the config file if you import `i3pystatus.core.util.internet` or by adding an optional argument to Status class.
``` python
class internet:

    address = ("google-public-dns-a.google.com", 53)

    def __new__(self):
        try:
            socket.create_connection(self.address, 1).close()
            return True
        except OSError:
            return False
```

enkore:
#280 merged, implementing this

PR #522, opened 2017-01-11
core: require internet is now updating each 5 seconds

hasB4K:
Hi,

When I updated my PR #521 I noticed that `require(internet)` is not working as expected, and is only runned once for each module (because of the usage of __new__)...
I think that `require(internet)` should update his status globally at a fixed interval (interval that I have set to 5 seconds).
This PR fix that.

Regards,

##
Mathis 'hasB4K' Felardos

### DIFF
--- a/i3pystatus/core/util.py
+++ b/i3pystatus/core/util.py
@@ -1,9 +1,10 @@
 import collections
 import functools
+import inspect
 import re
 import socket
 import string
-import inspect
+import time
 from threading import Timer, RLock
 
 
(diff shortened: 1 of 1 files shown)


hasB4K:
I think this PR is now fully functional :)

richese:
I think it works just fine as it is. 
I have added some log messages to test it and it gets called every time like it should (tested with external_ip and updates modules).
Are you sure it is not caused by your network like in #234 ? 

```python
import logging

class internet:
    address = ("google-public-dns-a.google.com", 53)

    def __new__(cls):
        try:
            socket.create_connection(cls.address, 1).close()
            logging.getLogger('i3pystatus').info('internet true')
            return True
        except (OSError, socket.gaierror):
            logging.getLogger('i3pystatus').info('internet false')
            return False
```

hasB4K:
You are right @richese. You made me realise that the issue that I have should not even occur...
However this is not caused by my network. Somehow the default `require(internet)` doesn't work properly on my machine, but my PR does. I need to investigate this later. This is weird!

enkore:
"class internet: __new__(): ... return True/False" is pretty much a hacky way of making a function "internet" that also can have that "address" attribute.

It'd be better to just write it as such:

```py
def internet():
        try:
            socket.create_connection(internet.address, 1).close()
            return True
        except (OSError, socket.gaierror):
            return False

internet.address = ("google-public-dns-a.google.com", 53)
```

hasB4K:
Yeah, indeed @enkore, I know that :). I think I'm just too tired today, and I tested it wrong...
Somehow I thought that the predicate was only instantiate once... Added to that an impatient test, you get the wrong conclusion...

I just tested it again, the PR #521 works fine without the #522... (obviously). So I think that the `external_ip` module can be added safely!

Thank you again to you two, @enkore and @richese!

PR #525, opened 2017-01-13
Remove playerctl module

m45t3r:
Playerctl uses the exactly same interface as now_playing, however it depends on a external binary dependency instead of dbus-python. And now_playing is more feature complete and follows the standard set by other modules (like the mpd module), while playerctl used an alien configuration.

Spotify module used to depend on playerctl. Actually, playerctl was the whole spotify module. So now it uses now_playing. I don't know if spotify only works with playerctl, however in theory if depends on the same dbus interface implemented on now_playing. Needs testing.

Fix some typos in redshift module documentation too.


ncoop:
Installed `spotify` from AUR to test this. It works with default configuration.

My concern is there's an i3pystatus user out there on NetBSD or something who has playerctl, but not dbus. Seems unlikely.

enkore:
Thanks for testing

m45t3r:
@ncoop FYI, playerctl used the same DBUS interface.

Issue #523, opened 2017-01-12
Xkblayout: AssertionError


m45t3r:
Xkblayout module seems broken after #492:

```
2017-01-12 00:50:05,471 [ERROR   ][i3pystatus.xkblayout.Xkblayout 76] Exception in Thread-3 at qui 12 jan 2017 00:50:05 BRST, module Xkblayout
Traceback (most recent call last):
  File "/home/m45t3r/.dotfiles/i3/.config/i3pystatus/venv/src/i3pystatus/i3pystatus/core/threading.py", line 68, in __call__
    self.workload()
  File "/home/m45t3r/.dotfiles/i3/.config/i3pystatus/venv/src/i3pystatus/i3pystatus/core/modules.py", line 285, in __call__
    self.run()
  File "/home/m45t3r/.dotfiles/i3/.config/i3pystatus/venv/src/i3pystatus/i3pystatus/xkblayout.py", line 79, in run
    "symbol": self._xkb.group_symbol,
  File "/home/m45t3r/.dotfiles/i3/.config/i3pystatus/venv/lib/python3.6/site-packages/xkbgroup/core.py", line 338, in group_symbol
    s_mapping = {symdata.index: symdata.symbol for symdata in self._symboldata_list}
  File "/home/m45t3r/.dotfiles/i3/.config/i3pystatus/venv/lib/python3.6/site-packages/xkbgroup/core.py", line 428, in _symboldata_list
    return _parse_symbols(b_symbol_str.decode(), ["pc", "inet", "group"])
  File "/home/m45t3r/.dotfiles/i3/.config/i3pystatus/venv/lib/python3.6/site-packages/xkbgroup/core.py", line 486, in _parse_symbols
    assert len(indices) == len(set(indices))    # No doubles
AssertionError
```

Problm seems to be in https://github.com/enkore/i3pystatus/blob/master/i3pystatus/xkblayout.py#L79. In fact, trying in terminal:

```
Python 3.6.0 (default, Dec 24 2016, 08:03:08) 
[GCC 6.2.1 20160830] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from xkbgroup import XKeyboard
>>> xkb = XKeyboard()
>>> xkb.group_symbol
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #527, opened 2017-01-14
Forcibly update xkblayout module internal state

m45t3r:
So when the user modifies the layout externally (for example, using setkbmap) the update is correctly reflect in i3pystatus.

Fix issue #523.

### DIFF
--- a/i3pystatus/xkblayout.py
+++ b/i3pystatus/xkblayout.py
@@ -72,7 +72,14 @@ class Xkblayout(IntervalModule):
     def change_layout(self, increment=1):
         self._xkb.group_num += increment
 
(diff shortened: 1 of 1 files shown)


m45t3r:
Sorry, bad idea to open PR without more testing.

Now I get `Xkblayout: X11Error: Cannon open display ".0".` after i3pystatus run for some time, probably because this PR is doing so many requests to X11 that X11 starts to block requests. Strangely enough, just restarting i3pystatus seems to fix, so maybe it is another problem altogether. And I still can't change layouts after changing the layout externally, it used to work without the xkbgroup dependency.

PR #528, opened 2017-01-14
Revert "Forcibly update xkblayout module internal state"

m45t3r:
Sorry, bad idea to open PR without more testing.

Now I get Xkblayout: X11Error: Cannon open display ".0". after i3pystatus run for some time, probably because this PR is doing so many requests to X11 that X11 starts to block requests.

And I still can't change layouts after changing the layout externally, it used to work without the xkbgroup dependency.

Going to open another issue, maybe @hcpl there is a better idea on how to fix those bugs.

### DIFF
--- a/i3pystatus/xkblayout.py
+++ b/i3pystatus/xkblayout.py
@@ -72,14 +72,7 @@ class Xkblayout(IntervalModule):
     def change_layout(self, increment=1):
         self._xkb.group_num += increment
 
-    def update_state(self):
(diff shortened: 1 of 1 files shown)


hcpl:
@m45t3r, the reason I added the xkbgroup dependency was the desire to have a finer-grained control over current xkb layouts than `setxkbmap` could provide.

Yeah, any new way should extend, not reduce the available functionality. So sorry again for an imperfect implementation :( I'm open for suggestions!

PR #530, opened 2017-01-15
taskwarrior: Add `project` formatter for next task project

Xenopathic:
It's useful to be able to see what project the next task belongs to.

I thought about some other potential formatters too like due date or tags, but couldn't think of a sensible way to parse that data (e.g. pretty printing the number of days to a due date) so I've only added the project formatter. As I start to use this plugin more I might consider adding those other formatters, if the need arises.


enkore:
[borg](https://github.com/borgbackup/borg/blob/master/src/borg/helpers.py#L729) has a useful function for formatting time-deltas, although for the bar one might want a more concise format.

Xenopathic:
@enkore Ooh, borg looks perfect. If I do implement that it will be in a separate PR though, not in this one.

PR #532, opened 2017-01-15
taskwarrior: Prevent KeyError on missing project

Xenopathic:
Slight error on my part in the original PR, should have been this fallback to prevent errors.

### DIFF
--- a/i3pystatus/taskwarrior.py
+++ b/i3pystatus/taskwarrior.py
@@ -80,7 +80,8 @@ class Taskwarrior(IntervalModule):
 
         if self.next_task is not None:
             format_values['next'] = self.next_task['description']
-            format_values['project'] = self.next_task['project']
+            format_values['project'] = self.next_task['project'] \
+                    if 'project' in self.next_task else ''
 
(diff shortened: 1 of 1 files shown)


i3pystatus/taskwarrior.py
```diff
@@ -80,7 +80,8 @@ def run(self):
 
         if self.next_task is not None:
             format_values['next'] = self.next_task['description']
-            format_values['project'] = self.next_task['project']
+            format_values['project'] = (self.next_task['project'] if 'project' in self.next_task
+                                        else '')
```
facetoe: It might be nicer to do:

```python
self.next_task.get('project', '')
```

https://docs.python.org/3.1/library/stdtypes.html#dict.get

Xenopathic: I did not know this was a thing. You've just revolutionised my Python coding!

facetoe: Yep I was happy when I found out about it as well, much cleaner than the alternatives!

PR #533, opened 2017-01-15
Rename module: now_playing to mpris.

ncoop:
The name is more appropriate, but this will break existing configs.


m45t3r:
While I think this is a better name for this module, I think breaking the API just to make this change is too much.

facetoe:
I'd have to agree with @m45t3r on this one. It's a better name but would break many configs as I'm guessing this is one of the more popular modules. I'm going to go ahead and close this PR but I'm happy to reopen it if anyone has any objections.

PR #543, opened 2017-02-05
Adding audacious, next and previous commands using audtool.

mohamed-aziz:
Why is next and previous not implemented in now_playing module?


richese:
Because you can do it by using the generic `player_command` callback?
```python
on_leftclick = ["player_command", "Next"]
on_leftclick = ["player_command", "Previous"]
```
I think the module documentation is clear enough.

mohamed-aziz:
@richese right, sorry for the inconvenience.

PR #548, opened 2017-02-23
codestyle -E722

enkore:


ncoop:
E722 is one of the good ones, I think.

Can Travis be made to grandfather the existing violations instead?

PR #551, opened 2017-02-28
Wildcard interface names

klloe:
allows you to use wildcards such as `wlp` or `enp` in the config. useful if you use the same config on multiple hosts to get the first (wireless) device. kinda like [__first__](https://i3wm.org/i3status/manpage.html#_wireless) does

### DIFF
--- a/i3pystatus/network.py
+++ b/i3pystatus/network.py
@@ -327,6 +327,12 @@ class Network(IntervalModule, ColorRangeModule):
         else:
             get_wifi_info = False
 
+        # wildcard interface names
+        for s in netifaces.interfaces():
+            if self.interface in s:
+                self.interface = s
+                break
+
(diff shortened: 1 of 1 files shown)


facetoe:
Would it be worth having a wildcard syntax for this? That way this code will only be executed for those that really want it/ Something like:

```python
if '*' in self.interface:
    ... apply wildcard logic
```

enkore:
https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch

facetoe:
Oh nice, yes that's perfect.

m45t3r:
There is glob too: https://docs.python.org/3/library/glob.html?highlight=glob#module-glob

facetoe:
@klloe does #562, solve the same problem as this PR?

klloe:
@facetoe nice job, yes I think so! Didn't try yet tho.

PR #552, opened 2017-03-01
pulseaudio: hides some error messages

sileht:
When we move the sink of an application, it can fail, then pacmd
print error message on stdout, breaking the i3pystatus json.

This changes hides these messages since their are not really usefull.


facetoe:
Nice.

PR #559, opened 2017-03-30
Another fix for Weather.com API changes

terminalmage:
It appears that the page data loads in one of two ways. I haven't
figured out what causes either, but this makes it so that we detect both
methods and load the relevant data either way.

### DIFF
--- a/i3pystatus/weather/weathercom.py
+++ b/i3pystatus/weather/weathercom.py
@@ -40,25 +40,79 @@ class WeathercomHTMLParser(HTMLParser):
                 exc_info=True
             )
 
+    def load_json(self, json_input):
+        self.logger.debug('Loading the following data as JSON: %s', json_input)
+        try:
+            return json.loads(json_string)
+        except json.decoder.JSONDecodeError as exc:
+            self.logger.debug('Error loading JSON: %s', exc)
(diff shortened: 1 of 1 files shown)


CPrompt:
pulled this merge and it doesn't error on the screen but it doesn't load any data.  Log shows this:

  File "/usr/lib/python3.5/site-packages/i3pystatus/weather/weathercom.py", line 36, in get_weather_data
    self.feed(html)
  File "/usr/lib64/python3.5/html/parser.py", line 111, in feed
    self.goahead(0)
  File "/usr/lib64/python3.5/html/parser.py", line 165, in goahead
    self.handle_data(rawdata[i:j])
  File "/usr/lib/python3.5/site-packages/i3pystatus/weather/weathercom.py", line 112, in handle_data
    weather_data = self.load_json(line.split('=', 1)[1].lstrip())
  File "/usr/lib/python3.5/site-packages/i3pystatus/weather/weathercom.py", line 46, in load_json
    return json.loads(json_string)
NameError: name 'json_string' is not defined

terminalmage:
@CPrompt I made two separate PRs to fix this issue. If you just pulled this one commit, then I'm not surprised you're getting an error.

terminalmage:
#558 is the other one.

PR #561, opened 2017-04-13
Fix weathercom module

drwahl:
The weathcom module was broken due to referencing a variable that did
not exist.  It appears this was a typo or copy/paste error.  Either way,
this patch fixes that error and the weathcom module works again!


terminalmage:
Ahh, thanks. Yes, I broke this when I was adjusting the Weather.com backend in https://github.com/enkore/i3pystatus/pull/558 and https://github.com/enkore/i3pystatus/pull/559 to reflect changes in Weather.com. Thanks!

PR #562, opened 2017-04-15
Add option for detecting active interface.

facetoe:


Glumanda99:
@facetoe I just tried this option and it is not what I was looking for. If I use `detect_active` on both network status items e.g. ethernet and wireless. I just get the same interface from both. I guess I'll just compare the hostname and use a different config on each device.

PR #563, opened 2017-04-16
Calculate battery percentage from energy instead of average of batteries

pajowu:
Average only works if all batteries have the same max energy level. If setups with different sizes the smaller ones influence the percentage overproportianly strong

### DIFF
--- a/i3pystatus/battery.py
+++ b/i3pystatus/battery.py
@@ -78,6 +78,9 @@ class BatteryCharge(Battery):
     def wh_remaining(self):
         return self.battery_info['CHARGE_NOW'] * self.battery_info['VOLTAGE_NOW']
 
+    def wh_total(self):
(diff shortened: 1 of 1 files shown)


facetoe:
Looks like this PR broke the test - https://travis-ci.org/enkore/i3pystatus/builds/222632424.

pajowu:
It seems like some batteries report the charge, while some report the energy. These are different units according to [[1]](https://bbs.archlinux.org/viewtopic.php?id=131025). I am not sure how easy it would be to convert them in case the script meets a device where different batteries use different units. My suggestions is to use this method in case all batteries report either energy or charge (I assume in that cases it can be mixed) and use the old, inaccurate way if they are not the same.

pajowu:
On second looks, it quite easy to convert the units. I fixed that in the second commit. Now it should also work on systems with batteries that report different units

facetoe:
Awesome, looks good.

Issue #564, opened 2017-04-20
Show different string in network module depending on active interface type


rscholer:
I'm testing the new "detect_active" function in the network module.

What I would like to do is to print "Ethernet" when my ethernet connection is active, and "Wifi", when my wireless connection is active.

Can this already be done, or is it necessary to implement this functionality in the module?

facetoe:
This may be possible using the [callback](https://i3pystatus.readthedocs.io/en/latest/configuration.html#callbacks) functionality. If you know the name of your Ethernet and Wifi interfaces you could probably write a Python function to modify the output accordingly. If this is a feature that people want it would also be easy to add a a configurable mapping to the module.

PR #565, opened 2017-04-21
Use different format strings for each auto-detected network interface

rscholer:
The new setting `format_active_up`, which is a dictionary, will be used if `detect_active` is enabled.

Each key is a valid [fnmatch](https://docs.python.org/3.5/library/fnmatch.html?highlight=fnmatch#module-fnmatch) glob pattern.

If no match is found, `format_up` is used instead.

This pull request also includes a small fix, which disables the calculation of color values if `detect_active` is enabled and `dynamic_color` is disabled. This is necessary, because `color_up` was not correctly used.

Closed issues: #564.

### DIFF
--- a/i3pystatus/network.py
+++ b/i3pystatus/network.py
@@ -417,24 +417,27 @@ class Network(IntervalModule, ColorRangeModule):
             format_values['network_graph'] = self.get_network_graph(kbs, limit)
             format_values['kbs'] = "{0:.1f}".format(round(kbs, 2))
 
-            if self.separate_color and self.pango_enabled:
-                color = self.color_up
-                color_template = "<span color=\"{}\">{}</span>"
-                per_recv = network_usage["bytes_recv"] * self.divisor / (self.recv_limit * 1024)
-                per_sent = network_usage["bytes_sent"] * self.divisor / (self.sent_limit * 1024)
-                c_recv = self.get_gradient(int(per_recv * 100), self.colors, 100)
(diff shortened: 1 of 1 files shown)


i3pystatus/network.py
```diff
@@ -410,31 +417,39 @@ def run(self):
...
         else:
             color = None
 
         if sysfs_interface_up(self.interface, self.unknown_up):
             if not color:
                 color = self.color_up
             format_str = self.format_up
+
+            if self.detect_active:
+                for pattern in self.format_active_up:
```
facetoe: Do you think it makes sense to sort the keys and have a break if there is a match? That way the order of evaluation is always the same and it will always set the format_string to the first match.

rscholer: > order of evaluation is always the same

Most people probably will only define a pretty limited amount of patterns, so that the 
penalty of sorting the keys is probably equal to iterating over unused patterns or even worse. 

The user can use `collections.OrderedDict`, if a permanent evaluation order is required, or the amount of non-matched patterns is to large.

> and have a break if there is a match

I thought about adding a break statement at first, but then I realized that the user could define quite clever rules without it:
```python
format_active_up = collections.dictionary(
                                     [
                                         ('eth*', 'Ethernet'), 
                                         ('eth0', 'First ethernet'),
                                     ]
                                 )
```

facetoe: Yep, makes sense. It's more flexible this way.

PR #567, opened 2017-04-26
Update openvpn

erfanio:
openvpn 2.4 has changed their services API so the clients are called openvpn-client now

### DIFF
--- a/i3pystatus/openvpn.py
+++ b/i3pystatus/openvpn.py
@@ -56,7 +56,6 @@ class OpenVPN(IntervalModule):
     def init(self):
         if not self.vpn_name:
             raise Exception("vpn_name is required")
-                # use new service names if openvpn version is 2.4 (2.5 is not out yet...)
 
         if self.openvpn_lastest:
             self.status_command = "bash -c 'systemctl show openvpn-client@%(vpn_name)s | grep ActiveState=active'"
--- a/i3pystatus/openvpn.py
+++ b/i3pystatus/openvpn.py
@@ -47,7 +47,7 @@ class OpenVPN(IntervalModule):
         ("status_down", "Symbol to display when down"),
         ("status_up", "Symbol to display when up"),
         ("vpn_name", "Name of VPN"),
-        ("openvpn_lastest", "Use lastest openvpn service names"),
+        ("use_new_service_name", "Use lastest openvpn service names"),
         ("vpn_up_command", "Command to bring up the VPN - default requires editing /etc/sudoers"),
         ("vpn_down_command", "Command to bring up the VPN - default requires editing /etc/sudoers"),
         ("status_command", "command to find out if the VPN is active"),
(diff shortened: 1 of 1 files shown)


erfanio:
Just saw the comment about being aware of this issue
I pushed a new commit that automatically detects that if openvpn version is 2.4, and then uses the new service names

i3pystatus/openvpn.py
```diff
@@ -54,7 +54,13 @@ class OpenVPN(IntervalModule):
     def init(self):
         if not self.vpn_name:
             raise Exception("vpn_name is required")
-
+        
+        # use new service names if openvpn version is 2.4 (2.5 is not out yet...)
+        if run_through_shell('openvpn --version | grep -i "openvpn 2.4"', True).out.strip():
```
facetoe: This is a good idea, but I think it should be an option. People might set `vpn_up_command` and `vpn_down_command` to execute scripts instead of openvpn directly and this would clobber their configuration.

erfanio: that's true, I didn't think about that.
Do you think it's worth checking if the command is not set, then override it?
like `if self.vpn_down_command == <insert old value>`
but this is super messy

facetoe: You could add it as a module paramater with a default of false, that way you won't break existing configs and people can still use the feature.

erfanio:
I changed it to use a config.
I could add a variable for the service name that is injected into the commands, but that might make it more confusing

i3pystatus/openvpn.py
```diff
@@ -46,6 +47,7 @@ class OpenVPN(IntervalModule):
         ("status_down", "Symbol to display when down"),
         ("status_up", "Symbol to display when up"),
         ("vpn_name", "Name of VPN"),
+        ("openvpn_lastest", "Use lastest openvpn service names"),
```
facetoe: Looks good, and this is a bit picky, but do think something like `detect_service_syntax` might be a better name?

erfanio: I don't know what `detect_service_syntax` is

erfanio: oooh do you mean just name the variable `detect_service_syntax`?

facetoe: Fair enough, not the best name then! Maybe `openvpn_latest`? Lastest isn't commonly used to describe the most recent.

facetoe: Sorry I didn't see your reply, yes for the variable name. Either that or  something like `openvpn_latest` would be better I think.

erfanio: how about `use_new_service_name`?

facetoe: Yep, sounds good!

erfanio:
You should probably squash the commits 😅

i3pystatus/openvpn.py
facetoe: Oops, I think you forgot to change it here too :)

facetoe:
Thanks!

PR #569, opened 2017-05-09
space api module

klloe:
[Space API ](http://spaceapi.net/) module to show if hackerspace is open

Example:
```
from i3pystatus import Status

status = Status()

status.register("spaceapi",
                url="https://it-syndikat.org/api/status.php",
                format="S: {state}")

status.run()
```

### DIFF
--- a/i3pystatus/spaceapi.py
+++ b/i3pystatus/spaceapi.py
@@ -1,0 +1,48 @@
+import json
+
+from urllib.request import urlopen
+from datetime import datetime
+
+from i3pystatus import IntervalModule
+from i3pystatus.core.util import internet, require
+
+
+class SpaceAPI(IntervalModule):
+    """
+    Show if a hackerspace is open
+
+    .. rubric:: Available formatters
+
+    * {state}
+    * {message}
+    * {lastchange}
+    """
+
+    data = {}
+    format = "S: {state}"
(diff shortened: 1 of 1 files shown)


i3pystatus/spaceapi.py
```diff
@@ -0,0 +1,48 @@
...
+        ("url", "spaceapi endpoint"),
+        ("format", "format string used for output."),
+        ("color_open", "color if hackerspace is opened"),
+        ("color_closed", "color if hackerspace is closed"),
+        ("interval", "update interval")
+    )
+
+    @require(internet)
+    def run(self):
+        res = urlopen(self.url)
```
facetoe: The call to `urlopen` can fail if there are issues with the connection, so it's best to handle that case.

i3pystatus/spaceapi.py
```diff
@@ -0,0 +1,48 @@
...
+        api = json.loads(res.read())
```
facetoe: The read call can also fail. It might be an idea to wrap these two lines in some exception handling to ensure the module doesn't error out if the connection becomes unreliable.

facetoe:
Looks good! I added a couple of notes about exception handling but apart from that looks good to merge.

klloe:
@facetoe :heart_eyes_cat:

Issue #577, opened 2017-06-10
Weather data is inaccurate: seems like degrees F and C are being confused with one another


fusion809:
Hi,

I have this in my config script:

```python
status.register(
    'weather',
    format='{condition} {current_temp}{temp_unit}',
    colorize=True,
    hints={'markup': 'pango'},
    backend=weathercom.Weathercom(
        location_code='ASXX0117:1:AS',
        units='metric',
    ),
)
```

and it's giving a ridiculously high temperature of 70&deg;C even though in the place this location code pertains to (Townsville, Australia) the temperature is shown [here](https://weather.com/en-AU/weather/today/l/ASXX0117:1:AS) as just 21&deg;C at this point of time. My config script is [here](https://pastebin.com/K2FdHr3f). 

Thanks for your time,
Brenton

fusion809:
Hmm this issue seems to only occur sometimes not all the time. Now it's back to saying "Fair 22&deg;C".

facetoe:
It doesn't look like the module does any conversion between fahrenheit and celcius.

Do you still get the results in fahrenheit?

fusion809:
Sure do. Getting it right now, here's a screenshot:

![screenshot_20170612_200726](https://user-images.githubusercontent.com/4717341/27029240-ccd6f768-4faa-11e7-8a4a-d06a3498a13b.png)

facetoe:
How often have you seen it correctly reporting Celcius?

fusion809:
If I had to hazard a guess I'd say about half, maybe less of the time I've run i3pystatus. I switch between i3pystatus and C/C++/Shell-based i3 config.

facetoe:
Can you test #582 and see if it resolves your issue? You will need to modify your config as follows:

```python
status.register(
    'weather',
    format='{condition} {current_temp}{temp_unit}',
    colorize=True,
    hints={'markup': 'pango'},
    backend=Weathercom(
        location_code='ASXX0117:1:AS',
        locale='en-AU',
        units='metric',
    ),
)
```

fusion809:
Well I'll be, it works! Thanks

facetoe:
No worries. I'll merge the PR in a few days to give others time to comment.

PR #582, opened 2017-06-12
weathercom.py: Add option to specify locale.

facetoe:
It appears weather.com returns values in fahrenheit by default. This PR allows specifiying a locale that will cause weather.com to return values in units standard for that country (eg, en-AU==Celcius).

Potential fix for #577

### DIFF
--- a/i3pystatus/weather/weathercom.py
+++ b/i3pystatus/weather/weathercom.py
@@ -120,9 +120,6 @@ class Weathercom(WeatherBackend):
     forecast page, the code you need will be everything after the last slash in
     the URL (e.g. ``94107:4:US``).
 
-    To receive values in units standard to your country, pass the relevant locale.
-    For example, locale='en_AU'.
-
     .. _weather-usage-weathercom:
 
     .. rubric:: Usage example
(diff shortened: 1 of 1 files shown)


terminalmage:
This should already be handled by the ``units`` param. It defaults to metric, so we don't need to add a separate argument for this, we just need to handle the URL differently when units == metric.

terminalmage:
Note that the Weather.com backend was recently re-done because Weather.com changed their API. You used to be able to just specify the units in the queryargs, That may no longer be possible though, so we could just pick a known metric locale and use that when units == metric.

facetoe:
Yep, fair point. I guess we could just set the "locale" to en-AU when units==metric. That should return results in Celcius etc.

terminalmage:
Yeah, that would do it.

When Weather.com changed their API, they removed the old JSON feed and the weather data is now fetched by some script and dumped into the page source as a blob of javascript. So I had to rewrite the module to parse the HTML instead.

I should have checked the metric units when I did the rewrite, shame on me for being so America-centric. :smile:

facetoe:
Haha all good, I've changed it to determine the locale based on `units`.

i3pystatus/weather/weathercom.py
```diff
@@ -123,6 +120,9 @@ class Weathercom(WeatherBackend):
     forecast page, the code you need will be everything after the last slash in
     the URL (e.g. ``94107:4:US``).
 
+    To receive values in units standard to your country, pass the relevant locale.
+    For example, locale='en_AU'.
+
```
terminalmage: This no longer has to be here since the ``locale`` option has been removed.

facetoe: Good catch!

terminalmage:
LGTM

PR #585, opened 2017-06-13
Openvpn: modularize and add conffile backend

drwahl:
This patchset moves the openvpn module to a backend-based module, converts the original systemd module to a backend, and adds a config file based backend.  This will allow users who run openvpn outside of systemd to get status of their client(s).


facetoe:
Nice work @drwahl!

i3pystatus/openvpn/__init__.py
```diff
@@ -0,0 +1,53 @@
+from i3pystatus import IntervalModule
+
+
+class Openvpn(IntervalModule):
+    """
+    Generic OpenVPN manager
+
+    The `backend` setting determines the backends to use.
```
facetoe: A quick example would be useful here.

drwahl: Done and done!

i3pystatus/openvpn/systemd.py
```diff
@@ -1,12 +1,12 @@
-from i3pystatus import IntervalModule
+from i3pystatus.openvpn import Backend
```
facetoe: This throws an import error for me. Should `Systemd` subclass `SettingsBase` like the `File` module?

facetoe:
The `Systemd` backend has an issue where the check for `vpn_name` at https://github.com/drwahl/i3pystatus/blob/fafcbc17b683d946ef91b348ba0567a06f0a83eb/i3pystatus/openvpn/systemd.py#L41 always fails as `vpn_name` is no longer an attribute of the class.

i3pystatus/openvpn/systemd.py
drwahl: Yup, that was a copy/paste error during development. This should be addressed now.

facetoe:
Actually on second thought there is no reason to make `vpn_name` required as the `File` backend doesn't require it.

drwahl:
I'm having a hard time getting the formatting for sphinx correct.  I'm going to ignore that failure for the moment and try to come up with a solution for the `vpn_name` issue.

drwahl:
@facetoe Ok, this isn't a terribly graceful solution, but it is a solution nonetheless.  Give it a shot and let me know how it works.  You'll have to pass "name" to the systemd backend as well as "vpn_name" (to the initial module).

facetoe:
Thinking about this a bit more, maybe the "backend" pattern isn't the best one to use for this module. A simpler implementation might be to simply have a `type` (or similar) parameter and depending on whether that is `systemd` or `conffile` the relevant class could be initialized in the `init` method. This would ease configuration for the user and remove the issue with the `vpn_name` value as it could be passed to the `Systemd` constructor at initialization.

drwahl:
Your last comment inspired me to give this backend approach one more shot.  This latest patch does still use the backend approach, but it has a little more logic in the init to pass things into the backend, should it be needed.

If you still think the `backend` approach is the wrong way to address this, then I can work on whipping up a new patch using the `type` approach you suggested.

i3pystatus/openvpn/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+    format = "{vpn_name} {status}"
+    backend = None
+    connected = False
+    on_leftclick = ['toggle_connection']
+
+    def toggle_connection(self):
+        self.backend.toggle_connection()
+
+    def run(self):
+        if 'systemd.Systemd' in str(self.backend):
```
facetoe: Not really a fan of this, seems like there must be a better way. Also this code executes `backend.init()` each time `run` is called which is a bit weird, usually you expect `init` to be executed once on initialization.

facetoe:
This fails for me as `vpn_name` is not defined - https://github.com/enkore/i3pystatus/pull/585/files#diff-19da4f536f3b8a0945de418ba344a29fR42.

facetoe:
I feel like this can be cleaned up somewhat. Here are my suggestions:

* Rename the module from `openvpn` to `vpn`. Seeing as we are breaking backwards compatibility with this change and using the backend pattern we might as well make it more generic.

* Remove the requirement for `vpn_name` in the top level module. There is already a parameter `label` that can be used for display purposes and this eliminates the problem of injecting `vpn_name` into the Systemd backend or requiring it be defined twice.

* Allow backends to return a dictionary of formatters. This adds a bit of flexibility and will allow, for example, the Systemd backend to return the `vpn_name` formatter.

Let me know what you think.

facetoe:
@drwahl I had a go at implementing my suggestions above however I'm a bit stuck on how the conffile backend is supposed to work. The openvpn process is started with sudo, and is therefore running as either root or whatever user it drops privileges to, which means the `os.kill()` line would fail with a PermissionError. I can't think of how you could kill this process without writing a script to do it and allowing it in /etc/sudoers.

The other problem is that openvpn doesn't seem to clean up the pid file when it exits, which means it's possible to accidentally kill some random process that ends up with whatever pid is in that file. Seeing as the kill command most likely needs to be executed as root due to the permissions issue, that could be bad!

I'm guessing most people use either sysv init, upstart or systemd for interacting with openvpn. Supporting these three systems is fairly straight forward as it all boils down to the three commands used for starting, stopping and querying status. This could be achieved easily enough in the current module.

All in all, I just can't see a good argument for changing to the backend pattern. It doesn't seem to offer any advantages, makes configuration more complex, and breaks user's configs.

For these reasons I'm going to go ahead and close this PR for now. I'm happy to open it up again if you do some more work on it in the future.

drwahl:
Ya, I was noticing issues with the kill bit as well, but hadn't dug into that part too deep just yet.  Good catch with the pidfile being left around; that seems like an openvpn bug.  Thanks for the thorough feedback and effort though!

PR #589, opened 2017-06-25
Sabnzbd

pf56:
Displays the current status of [SABnzbd](https://sabnzbd.org/).


i3pystatus/sabnzbd.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def init(self):
+        """Initialize the URL used to connect to SABnzbd."""
+        self.url = self.url.format(host=self.host, port=self.port,
+                                   api_key=self.api_key)
+
+    def run(self):
+        """Connect to SABnzbd and get the data."""
+        try:
+            answer = urlopen(self.url + "&mode=queue").read().decode("UTF-8")
```
enkore: decode("UTF-8") -> UTF-8 is the default, you can write decode() with the same effect

i3pystatus/sabnzbd.py
```diff
@@ -0,0 +1,106 @@
...
+            answer = json.loads(answer)
+        except (HTTPError, URLError) as error:
```
enkore: None of the exceptions that "json.loads" could raise are caught here - move it out of the try: block?

i3pystatus/sabnzbd.py
```diff
@@ -0,0 +1,106 @@
...
+            self.output = {
+                "full_text": str(error.reason),
+                "color": "#FF0000"
+            }
+            return
+
+        # if answer["status"] exists and is False, an error occured
+        if (not answer.get("status", True)):
```
enkore: () are not needed here

i3pystatus/sabnzbd.py
```diff
@@ -0,0 +1,106 @@
...
+        if (self.is_paused()):
+            color = self.color_paused
+        elif (self.is_downloading()):
+            color = self.color_downloading
+        else:
+            color = self.color
+
+        self.output = {
+            "full_text": self.format.format(**queue) if (self.is_downloading())
+            else self.format_paused.format(**queue),
```
enkore: If you like you can pull this if/else out in a separate block - probably easier to read

PR #595, opened 2017-07-16
Reduce overhead of internet predicate.

facetoe:
* Use 8.8.8.8 instead of google-public-dns-a.google.com save a DNS lookup.

* Only check the connection at most once per second.

### DIFF
--- a/i3pystatus/core/util.py
+++ b/i3pystatus/core/util.py
@@ -359,8 +359,9 @@ class internet:
     """
     Checks for internet connection by connecting to a server.
 
-    Used server is determined by the `address` class variable which consists of
-    server host name/IP and port number.
+    Used server can be configured via the `address` class variable which expects a tuple in the form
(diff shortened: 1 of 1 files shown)


drwahl:
Can we make the "once per second" thing configurable? I'd personally prefer to have my setup be much less often.

i3pystatus/core/util.py
```diff
@@ -367,11 +369,23 @@ class internet:
         :py:func:`require`
 
     """
-    address = ("google-public-dns-a.google.com", 53)
+    address = ("8.8.8.8", 53)
+
+    last_checked = time.time()
```
enkore: Perhaps use a monotonic clock (perf_counter() or monotonic()) instead (time() can jump around arbitrarily, esp. near a machine reboot)

facetoe:
OK I've used time.perf_counter() for measuring elapsed time and added a `check_frequency` setting for controlling how long the connection status is cached for.

PR #596, opened 2017-07-25
Fix breakage introduced in #595

hobarrera:
PR #595 assumed a very specific network setup, and breaks the `internet` predicate on other setups. 

Noticeably, it broke on IPv6-only networks (including those using NAT64, and other modern transition mechanisms).

### DIFF
--- a/i3pystatus/core/util.py
+++ b/i3pystatus/core/util.py
@@ -370,7 +370,7 @@ class internet:
         :py:func:`require`
 
     """
-    address = ("8.8.8.8", 53)
+    address = ("google-public-dns-a.google.com", 53)
 
     check_frequency = 1
     last_checked = time.perf_counter() - check_frequency


facetoe:
Oops, apologies for the breakage. I guess a solution could be to cache the result of https://docs.python.org/3/library/socket.html#socket.getaddrinfo. That would preserve the previous behaviour but without needing to make constant DNS requests.

hobarrera:
I guess it makes sense to cache the result, but *if* connection fails, then retry the original address again.

I'm mostly thinking of a device switching between networks where the results would change (some captive portals that do DNS poisoning are probably the most common case).

facetoe:
OK I've submitted another PR to fix this (#597). Whenever the connection fails, query DNS and cache the result. From there on in as long as the connection is succeeding there is no need to query DNS again and everything should Just Work™.

facetoe:
Closing in favour of #597.

hobarrera:
Thanks for picking this up. I couldn't make the time to get around to completing this 😅

PR #600, opened 2017-08-02
Module for fetching and displays exchange rates with EXMO

fmarchenko:


i3pystatus/exmo.py
```diff
@@ -0,0 +1,95 @@
...
+            'https://exmo.me/ru/trade#?pair={}'.format(self.pair)
+        ]
+
+    def fetch_data(self):
+        response = requests.get('https://api.exmo.com/v1/ticker/')
+        return response.json()
+
+    @require(internet)
+    def run(self):
+        price_data = self.fetch_data().get(self.pair)
```
facetoe: This can throw an exception if there are any issues communicating with the API.

fmarchenko: Okay, I fix this problem

Issue #592, opened 2017-07-06
feature request: separate upload / download `kbs` and `network_graph`


cprn:
Very low priority but it would be nice to get `"{network_graph_down}{kbs_down} KB/s {network_graph_up}{kbs_up} KB/s"` in one format string instead of registering module two times with `graph_type='input'` and `graph_type='output'`.

cprn:
Awesome! Thank you!

PR #615, opened 2017-10-16
Separate network graphs for sent and received data.

schroeji:
This implements the feature requested in #592.
You can now use `{network_graph_recv}` and `{network_graph_sent}` to choose which graph you want to display and are no longer forced to register the module 2 times.
The `{kbs}` formatter has been removed because you can now use `{bytes_recv}` and `{bytes_sent}`
instead.

### DIFF
--- a/i3pystatus/network.py
+++ b/i3pystatus/network.py
@@ -435,7 +435,7 @@ class Network(IntervalModule, ColorRangeModule):
                         color = self.get_gradient(network_usage['bytes_recv'], self.colors, self.recv_limit)
                     elif self.coloring_type == "sent":
                         color = self.get_gradient(network_usage['bytes_sent'], self.colors, self.sent_limit)
-                    else :
+                    else:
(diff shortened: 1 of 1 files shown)


facetoe:
Looks good @schroeji, thanks!

MrBobBarker:
No more combined graph? I might just stick with the two graphs now, but some people still might want the combined feature. Also the default format_up is invalid now that {network_graph} has been split.
format_up = "{interface} {network_graph}{bytes_recv}KB/s"

facetoe:
Good catch with the default format @MrBobBarker, I've fixed that up in #616.

tyjak:
Agree with @MrBobBarker said, I don't have to much space on my tiny screen so combined graph was fined to me :) Now I had to choose beetween up or down graph :( Is this possible to bring back {network_graph} conbined ?
Thanks for the nice job anyway !

PR #617, opened 2017-11-02
New source of cpu information

qgnox:
Hi, this PR adds the following : 

Starting with kernel 4.13 /proc/cpuinfo only shows the base clock frequency of the processor, not the current. 

https://bugzilla.kernel.org/show_bug.cgi?id=197009#c1

So this change extracts the current frequency from /sys/devices/system/cpu and considers also if the cpu's are online or offline . Using only the "file" parameter override with **/sys** value: 

`status.register("cpu_freq", format="{avgg} GHz", file='/sys', interval=1)`

### DIFF
--- a/i3pystatus/cpu_freq.py
+++ b/i3pystatus/cpu_freq.py
@@ -33,18 +33,18 @@ class CpuFreq(IntervalModule):
             with open('/sys/devices/system/cpu/online') as f:
                 line = f.readline()
                 cpus_online = [int(cpu) for cpu in line.split(',') if cpu.find('-') < 0]
-                cpus_online_range = [cpu_range for cpu_range in line.split('-') if cpu_range.find('-') > 0 ]
+                cpus_online_range = [cpu_range for cpu_range in line.split('-') if cpu_range.find('-') > 0]
 
             for cpu_range in cpus_online_range:
-                cpus_online += [cpu for cpu in range(int(cpu_range[0]), int(cpu_range[1])+1)]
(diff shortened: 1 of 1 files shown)


facetoe:
When I run this I get the following error: `ValueError: max() arg is an empty sequence` which I think is being thrown here - https://github.com/enkore/i3pystatus/pull/617/files#diff-d285f3da6b6f43df736320cf10d287b2R41. 

System info:

``` bash
⇒ uname -r
4.13.10-1-ARCH

⇒ cat /sys/devices/system/cpu/online
0-7
```

I found these docs which might shed some light on the syntax of this file: https://www.kernel.org/doc/Documentation/cputopology.txt.

qgnox:
mmm I'm using it since the PR, testing in Archlinux vanilla, ck and zen 4.13.11: 
```
$i3pystatus -c i3pystatus-config-cpu.py
{"version": 1, "click_events": true}
[
[{"full_text": "\uf109 3216.819 GHz", "format": "\uf109 {avg} GHz", "name": "i3pystatus.cpu_freq.CpuFreq", "instance": "140611344007744", "markup": "none"}]
...
[
[{"full_text": "\uf109 1.89 GHz", "format": "\uf109 {avgg} GHz", "name": "i3pystatus.cpu_freq.CpuFreq", "instance": "140241599364680", "markup": "none"}]
```

Looks like what you execute doesnt have the last commit : e0d6a80

facetoe:
Ahh, you're right, I didn't have the last commit that fixed exactly this issue :)

PR #621, opened 2017-11-09
Module for monitoring miners speed on NiceHash.com

fmarchenko:


facetoe:
Did you only mean to create a PR for the `NiceHashSpeed` module? If so could you clean up the PR to only include this module? Makes it much easier to review.

If you want to submit all the modules in this PR it would be cool if you could split them up into separate PRs.

Cheers!

PR #624, opened 2017-11-19
temp: Ignore pysensors exceptions on unavailable subfeature

kyrias:
Sometimes a temperature sensor is detected, but its value isn't available, which causes pysensors to throw an exception.  This shouldn't cause the bar to break, so ignore any subfeature exceptions.

### DIFF
--- a/i3pystatus/temp.py
+++ b/i3pystatus/temp.py
@@ -42,12 +42,15 @@ def get_sensors():
     for chip in sensors.get_detected_chips():
         for feature in chip.get_features():
             if feature.type == sensors.FEATURE_TEMP:
-                name = chip.get_label(feature)
-                max = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_MAX)
-                current = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_INPUT)
-                critical = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_CRIT)
-                if critical:
-                    found_sensors.append(Sensor(name=name, current=current, maximum=max, critical=critical))
+                try:
+                    name = chip.get_label(feature)
+                    max = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_MAX)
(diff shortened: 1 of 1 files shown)


i3pystatus/temp.py
```diff
@@ -42,12 +42,15 @@ def get_subfeature_value(feature, subfeature_type):
...
-                if critical:
-                    found_sensors.append(Sensor(name=name, current=current, maximum=max, critical=critical))
+                try:
+                    name = chip.get_label(feature)
+                    max = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_MAX)
+                    current = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_INPUT)
+                    critical = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_CRIT)
+                    if critical:
+                        found_sensors.append(Sensor(name=name, current=current, maximum=max, critical=critical))
+                except:
```
facetoe: Is it possible to be more specific in the exception you catch?

kyrias: Done

i3pystatus/temp.py
```diff
@@ -42,12 +42,15 @@ def get_subfeature_value(feature, subfeature_type):
...
-                if critical:
-                    found_sensors.append(Sensor(name=name, current=current, maximum=max, critical=critical))
+                try:
+                    name = chip.get_label(feature)
+                    max = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_MAX)
+                    current = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_INPUT)
+                    critical = get_subfeature_value(feature, sensors.SUBFEATURE_TEMP_CRIT)
+                    if critical:
+                        found_sensors.append(Sensor(name=name, current=current, maximum=max, critical=critical))
+                except sensors.SensorsException
```
facetoe: Looks like you're missing a `:`.

kyrias: Ah, damned, fixed it but forgot to commit it.  Thanks.

PR #625, opened 2017-11-19
temp: Add support for wrapping pango output in single color

kyrias:
Currently there is no way to color the characters in the format string,
like "°C" any color other than full white if pango output is enabled.
This feature adds another setting, wrapper_color, that will be used to
wrap the whole output string.

### DIFF
--- a/i3pystatus/temp.py
+++ b/i3pystatus/temp.py
@@ -209,7 +209,8 @@ class Temperature(IntervalModule, ColorRangeModule):
         data['temp'] = max((s.current for s in found_sensors))
         return {
             'full_text': self.format.format(**data),
-            'urgent': self.get_urgent(found_sensors)
+            'urgent': self.get_urgent(found_sensors),
+            'color': self.color if not self.dynamic_color else None,
         }
 
     def get_urgent(self, sensors):
@@ -230,9 +231,7 @@ class Temperature(IntervalModule, ColorRangeModule):
             percentage = self.percentage(sensor.current, sensor.critical)
             if self.dynamic_color:
                 color = self.colors[int(percentage)]
-            else:
-                color = self.color
-            return self.format_pango(color, current_val)
+                return self.format_pango(color, current_val)
         return current_val
 
     def format_sensor_bar(self, sensor):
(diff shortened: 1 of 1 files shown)


facetoe:
Can't you do this in the format string?

kyrias:
You could, but it seems a bit ugly to force people to write the HTML themselves in their config for things that you can easily do with just color settings for the rest of the modules.

facetoe:
It looks like in the case that pango is enabled and `dynamic_color` is False, the `color` setting is not honoured. Seeing as the `color` setting already exists, it's probably best to fix that up rather than adding a new setting. This could be done by adding:

```python
'color': self.color if not self.dynamic_color else None
```
to the dict returned by `get_output_sensors()`.

kyrias:
I mean, it is honored, it's just only used for each specific sensor value.  But I guess it would make sense to just have the whole thing wrapped in a span when `dynamic_color` isn't used, yeah.

i3pystatus/temp.py
```diff
@@ -151,13 +151,15 @@ class Temperature(IntervalModule, ColorRangeModule):
         ('urgent_on', 'whether to flag as urgent when temperature exceeds urgent value or critical value '
                       '(requires lm_sensors_enabled)'),
         ('dynamic_color', 'whether to set the color dynamically (overrides alert_color)'),
+        ('wrapper_color', 'wrapper color to use for the whole bar item if pango output is enabled'),
```
facetoe: This should be good to remove now.

i3pystatus/temp.py
```diff
@@ -151,13 +151,15 @@ class Temperature(IntervalModule, ColorRangeModule):
...
         "color",
         "file",
         "alert_temp",
         "alert_color",
     )
     format = "{temp} °C"
     color = "#FFFFFF"
+    wrapper_color = None
```
facetoe: Also this one

facetoe:
Looks good, just a couple of unused settings to remove and it's good to merge.

kyrias:
Thanks for having the patient to deal with the sleep-deprived idiot me.  :)

facetoe:
Haha all good, thanks for the PR!

PR #628, opened 2017-11-28
Module for monitoring miners speed on NiceHash.com

fmarchenko:
I'm closed previous PR and create new with only NiceHash module.
I will be happy with the positive decision :-)

### DIFF
--- a/i3pystatus/nicehash_speed.py
+++ b/i3pystatus/nicehash_speed.py
@@ -1,0 +1,140 @@
+import datetime
+import requests
+import re
+import time
+import json
+
+from i3pystatus import IntervalModule, formatp
+from i3pystatus.core.util import internet, require, user_open
+
+__author__ = "Fedor Marchenko"
+__email__ = "mfs90@mail.ru"
+__date__ = "Nov 09, 2017"
+
+
+class NiceHashSpeed(IntervalModule):
+    """Module for monitoring you miner statistic on nicehash.com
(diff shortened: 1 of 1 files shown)


i3pystatus/nicehash_speed.py
```diff
@@ -0,0 +1,140 @@
...
+
+    _prev_speed = 0.0
+    _prev_status = ''
+    _prev_color = '#FFFFFF'
+
+    _api_url = 'https://api.nicehash.com/api?method=stats.provider.ex&addr={addr}&from={from_unixtime}'
+    _api_query_interval = 3
+
+    def __init__(self, *args, **kwargs):
+        super(NiceHashSpeed, self).__init__(*args, **kwargs)
```
facetoe: It's better to use the provided `init` method in the base class than `__init__` as some housekeeping is done before the module is initialized.

i3pystatus/nicehash_speed.py
```diff
@@ -0,0 +1,140 @@
...
+            'open_something',
+            'https://www.nicehash.com/miner/{}'.format(self.addr)
+        ]
+
+    def fetch_data(self):
+        """Fetching statistic data from nicehash API for you BTC address
+
+        :return: Return JSON data with statistic information about miners
+        """
+        with open('/tmp/nhapi.json', 'w+') as f:
```
facetoe: Do you really need to write out the `fetch_date_time` to a file? You could just cache it in memory.

PR #635, opened 2018-02-05
gpu_usage module

gdunstone:
I added a gpu_usage module. its a little barebones, but we dont exactly get a huge amount of information from nvidia-smi.


facetoe:
I can't test but looks good to me.

PR #636, opened 2018-02-25
Add mail backend for Exchange account

forscher21:


i3pystatus/mail/ews.py
```diff
@@ -0,0 +1,74 @@
...
+                        credentials=credentials,
+                        autodiscover=True,
+                        access_type=exchangelib.DELEGATE)
+            yield
+        except EWS_EXCEPTIONS as e:
+            # NOTE(sileht): retry just once if the connection have been
+            # broken to ensure this is not a sporadic connection lost.
+            # Like wifi reconnect, sleep wake up
+            # Wait a bit when disconnection occurs to not hog the cpu
+            print(e)
```
facetoe: You should use the logger as this output will cause i3bar to fail on invalid JSON.

i3pystatus/mail/ews.py
```diff
@@ -0,0 +1,74 @@
+import exchangelib
+import contextlib
+import time
+
+from i3pystatus.mail import Backend
+
+
+EWS_EXCEPTIONS = (Exception)
```
facetoe: Is this necessary considering there is only one exception?

i3pystatus/mail/ews.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class EWS(Backend):
```
facetoe: No biggie, but a more descriptive class name would be good.

i3pystatus/mail/ews.py
enkore: Note that EWS_EXCEPTIONS is not a tuple here, because `(1) == 1` in Python.

PR #640, opened 2018-03-31
Add a color map for mpd states

EgZvor:


i3pystatus/mpd.py
```diff
@@ -175,9 +178,10 @@ def run(self):
                 fdict[key] = fdict[key][:shrink] + "…"
 
             full_text = formatp(self.format, **fdict).strip()
+        color_map = defaultdict(lambda: self.color, self.color_map)
```
enkore: clever

enkore:
LGTM, thanks!

PR #648, opened 2018-05-23
Allow color setting for Xkblayout indicator

torimus:
Simple fix - enable text color setting instead of hardcoded one.


torimus:
Battery.wh_total didn't comply to design/last state option, so percentage_design and bar_design format placeholders didn't reflect that choice. Patch fixes that issue.

PR #655, opened 2018-06-13
Add Sonos module

erikmansson:
A module for controlling and displaying information from Sonos speakers. Uses SoCo (can be installed from pip).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,7 +61,8 @@ MOCK_MODULES = [
     "requests.packages.urllib3.response",
     "lxml.etree",
     "requests.adapters",
-    "exchangelib"
+    "exchangelib",
+    "soco"
 ]
 
 for mod_name in MOCK_MODULES:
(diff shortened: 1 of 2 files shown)


erikmansson:
How can I prevent the import error from becoming an issue?

enkore:
https://i3pystatus.readthedocs.io/en/latest/module.html#handling-dependencies

> To allow automatic generation of the docs without having all requirements of every module installed mocks are used. To make this work simply add all modules of dependencies (so no standard library modules or modules provided by i3pystatus) you import to the MOCK_MODULES list in docs/conf.py. This needs to be the actual name of the imported module, so for example if you have from somepkg.mod import AClass, you need to add somepkg.mod to the list.

In your case this would be `soco`.

erikmansson:
@enkore Any thoughts? Improvements that can be made?

enkore:
Nope ;)

erikmansson:
Thanks :)

PR #656, opened 2018-07-09
i3pystatus.network: prefer non link-locale v4 addresses

Mic92:
Addresses in 169.254.0.0/16 are auto-assigned as of RFC 3927.
These can be negotiated via the zeroconf protocol (avahi-autoip or systemd-networkd).
Usually those addresses are not so interesting and the addresses given by dhcp should be shown instead.


enkore:
Thanks!

Issue #651, opened 2018-05-31
pycodestyle is crying like a baby again


enkore:
```
+python3 -mpycodestyle --ignore E501,E741,E305,E722 i3pystatus tests
i3pystatus/mpd.py:174:28: W504 line break after binary operator
i3pystatus/network.py:356:16: W504 line break after binary operator
i3pystatus/now_playing.py:163:45: W504 line break after binary operator
i3pystatus/now_playing.py:165:45: W504 line break after binary operator
i3pystatus/reddit.py:78:6: W605 invalid escape sequence '\{'
i3pystatus/reddit.py:78:19: W605 invalid escape sequence '\w'
i3pystatus/reddit.py:78:22: W605 invalid escape sequence '\}'
i3pystatus/reddit.py:79:6: W605 invalid escape sequence '\{'
i3pystatus/reddit.py:79:16: W605 invalid escape sequence '\w'
i3pystatus/reddit.py:79:19: W605 invalid escape sequence '\}'
i3pystatus/reddit.py:80:6: W605 invalid escape sequence '\{'
i3pystatus/reddit.py:80:21: W605 invalid escape sequence '\}'
i3pystatus/reddit.py:80:24: W605 invalid escape sequence '\{'
i3pystatus/reddit.py:80:36: W605 invalid escape sequence '\}'
i3pystatus/window_title.py:59:13: W504 line break after binary operator
i3pystatus/window_title.py:60:17: W504 line break after binary operator
i3pystatus/window_title.py:77:13: W504 line break after binary operator
i3pystatus/window_title.py:78:14: W504 line break after binary operator
i3pystatus/window_title.py:79:14: W504 line break after binary operator
i3pystatus/core/imputil.py:16:17: W504 line break after binary operator
i3pystatus/core/imputil.py:17:17: W504 line break after binary operator
i3pystatus/core/util.py:274:10: W605 invalid escape sequence '\['
i3pystatus/core/util.py:274:10: W605 invalid escape sequence '\]'
i3pystatus/pulseaudio/__init__.py:213:20: W605 invalid escape sequence '\s'
i3pystatus/pulseaudio/__init__.py:213:24: W605 invalid escape sequence '\d'
tests/test_core_util.py:192:19: W605 invalid escape sequence '\['
tests/test_core_util.py:192:27: W605 invalid escape sequence '\]'
```

Does someone know a saner alternative? One that doesn't decide to do start complaining about code it declared fine two days ago?

[...]

PR #659, opened 2018-07-31
fix vdsl2 usage, also add "used" formatter

chestm007:
im unable to test for compatibility with other connection types, but this is working for my VDSL2+ connection


facetoe:
Looks good. The PEP errors look related to https://github.com/enkore/i3pystatus/issues/651

PR #660, opened 2018-08-05
allow to use vertical bar in battery

tshirtman:


i3pystatus/battery.py
```diff
@@ -320,6 +320,8 @@ def run(self):
             "remaining": TimeWrapper(0, "%E%h:%M"),
             "bar": make_bar(self.percentage(batteries)),
             "bar_design": make_bar(self.percentage(batteries, design=True)),
+            "vertical_bar": make_vertical_bar(self.percentage(batteries)),
+            "vertical_bar_design": make_vertical_bar(self.percentage(batteries, design=True)),
```
enkore: Try `make_vertical_bar(int(self.percentage...))`, which should make it work.

PR #662, opened 2018-08-06
Fixes deprecated usage of Formatter.format.

verri:
Formatter.format arguments are now positional-only (Python 3.7).

Check [here](https://docs.python.org/3/library/string.html#string.Formatter.format).


enkore:
I don't think this is deprecated.

PR #671, opened 2018-10-22
Add humanize_remaining formatter

janza:
Allows for formating the time remaning to next event in _human readable_ format. Work only if `humanize` module is available.


enkore:
Can you add this new optional dependency to the docs of the calendar module? Otherwise people might not find out about this useful feature ;)

janza:
I've updated the docstring

PR #672, opened 2018-10-23
temp.py: Fixed bug related to dynamic colors

tigerjack:
When a custom alert_temp was given, for any temperature value above `alert_temp`
the module generated an `IndexError: string index out of range` error.

The problem was that for any temperature greater than `alert_temp`, the percentage returned from `self.percentage` was above 99. So, I simply fixed the error setting a maximum value for the percentage to 99.


enkore:
Nice find!

PR #674, opened 2018-10-28
Glyph based representation of numeric values

dmelliot:
Add `make_glyph` function to core utils (and update battery module to use it) 

A new function (in line with similar functions such as `make_bar`), `make_glyph` allows generating a glyph based representation of a numeric value.

This provides a way for modules to easily expose functionality to use a set of glyphs (e.g. battery level icons from FontAwesome) to represent a numeric value (e.g. battery level).


enkore:
Cool!

PR #677, opened 2018-11-23
Travisci

chestm007:
module to track current build status of travisci builds

### DIFF
--- a/i3pystatus/travisci.py
+++ b/i3pystatus/travisci.py
@@ -21,8 +21,31 @@ class TravisCI(IntervalModule):
     * `{repo_status}` - repository status
     * `{repo_name}` - repository name
     * `{repo_owner}` - repository owner
-    * `{last_build_finished}` - <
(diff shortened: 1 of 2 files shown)


i3pystatus/travisci.py
```diff
@@ -0,0 +1,115 @@
...
+        self.last_build_duration = None
+        self.last_build_finished = None
+        self.repo_owner, self.repo_name = self.repo_slug.split('/')
+
+    def _format_time(self, time):
+        _datetime = dateutil.parser.parse(time)
+        return _datetime.strftime(self.time_format)
+
+    def run(self):
+        if internet():
```
enkore: You can use `@require(internet)` instead

chestm007: was totally unaware, thats 100% cleaner! i'll push changes

enkore:
I like how well you documented this, a nice addition. Thanks!

PR #679, opened 2018-11-30
stop iinet from filling the bar with a connection error if internet is down

chestm007:
currently this module fills the bar with a stacktrace, this fixes that


i3pystatus/iinet.py
```diff
@@ -64,25 +65,26 @@ def valid_response(self, response):
         return "success" in response and response['success'] == 1
 
     def run(self):
```
enkore:
```suggestion
    @require(internet)
    def run(self):
```

instead

PR #685, opened 2018-12-12
Deluge

chestm007:
Module to view data on a running deluge daemon.
As deluge works in a client/server model regardless of whether or not the daemon is on the same machine as the client, this can be used for a local daemon.

I haven't done any solid benchmarks, but based on a cursory look in `htop` and `dstat` the performance impact of even grabbing all of the `libtorrent` stats doesn't seem too intensive (barely cracked 2% cpu on my ryzen7)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,6 +33,7 @@ MOCK_MODULES = [
     "requests.exceptions",
     "bs4",
     "dota2py",
+    'deluge-client',
     "novaclient",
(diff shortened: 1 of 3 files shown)


chestm007:
found a couple of last minute issues, its ready now

chestm007:
closing as there is some compatibility issues with deluge 1.3.15 - i'll reopen when ive fixed them

PR #690, opened 2018-12-20
Fix TypeError when not using network traffic in format string

terminalmage:
These formatters are initialized as strings, so when they're not used we don't use use the code that updates the network counters, and the result is that we attempt to divide an empty string by the divisor value (an integer), resulting in a TypeError.

This ignores these errors, letting these formatters remain as empty strings.


terminalmage:
https://github.com/enkore/i3pystatus/pull/694 includes a better fix for this, closing.

PR #691, opened 2018-12-20
Update default value for github status URL

terminalmage:
status.github.com has been deprecated, this updates the default URL


chestm007:
This only affects the webpage thats launched in `launch_status_url` so should be perfectly safe to merge

terminalmage:
Yes, and the `status_url` config option can be used to override it, if GitHub gets rid of the old status page and people don't want to upgrade i3pystatus. This just changes the default URL.

PR #707, opened 2019-01-29
Added TLP module

netvl:
This module allows for displaying the status of the TLP system (power management tools), which can be either AC, battery or unknown (in case TLP is not running, for example). It is a very simple module which reads from a particular file and interprets the value in it to display it with correct semantics.

### DIFF
--- a/i3pystatus/tlp.py
+++ b/i3pystatus/tlp.py
@@ -1,0 +1,54 @@
+from i3pystatus import IntervalModule
+
+
+class Tlp(IntervalModule):
+    """
+    Shows the current mode of TLP (Linux power management tool), either
+    battery, AC or unknown.
+
+    .. rubric:: Available formatters
+
+    * `{output}` - one of the strings configured through the `*_text` settings
+    """
+    last_pwr_file = "/run/tlp/last_pwr"
+    bat_color = "#00FF00"
+    ac_color = "#FFAA00"
+    na_color = "#FF0000"
+    bat_text = "BAT"
+    ac_text = "AC"
+    na_text = "N/A"
+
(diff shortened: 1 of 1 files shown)


i3pystatus/tlp.py
```diff
@@ -0,0 +1,56 @@
+from pathlib import Path
+
+from i3pystatus import IntervalModule
+
+class Tlp(IntervalModule):
```
netvl: Not sure about naming, maybe it should be `TLP`?

chestm007:
You'll have to list `pathlib` in `docs/conf.py`to get the tests to pass

Also, is there a strong reasoning to use `pathlib.Path` as opposed to `os.path` in this case?

netvl:
@chestm007 not really, it's just I thought that pathlib is the modern API for working with paths and files, so it is supposed to be used by default. I think I'll just change it to the regular `open` call then, to keep things simpler.

netvl:
The tests are still failing, but not because of my changes:
```
+python3 -mpycodestyle i3pystatus tests
i3pystatus/timewarrior.py:52:17: E117 over-indented
i3pystatus/timewarrior.py:54:17: E117 over-indented
i3pystatus/core/settings.py:118:17: E117 over-indented (comment)
i3pystatus/core/settings.py:119:17: E117 over-indented
i3pystatus/core/util.py:567:13: E117 over-indented
i3pystatus/core/util.py:569:13: E117 over-indented
```

chestm007:
hmm yeah, the lines being flagged have been like that for ages - maybe the linter changed upstream - i'll merge anyway.

PR #731, opened 2019-04-12
fixed broken percentage_design & extended testing

lucible:
The percentage definition in BatteryChecker uses the wh_remaining & wh_total methods instead of the percentage method. While battery.percentage takes into account the design=True/False flag, battery.wh_total does not. This lead to percentage and percentage_design being reported as exactly the same number (with percentage_design being incorrect). I have edited the two wh_total functions to recognize the design flag and return the proper value when it is true, and extended the testing to percentage & percentage_design.

Note that test_battery_issue729 will fail (reporting FULL rather than DIS) until #730 is merged.


chestm007:
closing and reopening to trigger a new build

PR #734, opened 2019-04-22
i derped and commited these :facepalm:

chestm007:
also added .idea/* to .gitignore so this doesn't happen again.


terminalmage:
I once accidentally committed an API key for Weather Underground to this repo. It happens :laughing:

chestm007:
hahaha gold!

PR #747, opened 2019-05-30
Some fixes to backlight module

thiagokokada:
This PR brings 2 fixes to `backlight` module:
- It fixes the following exception when the user does not have a backlight in the system (i.e. Desktop) and tries to adjust the brightness:
```
2019-05-27 23:05:53,754 [CRITICAL][i3pystatus.backlight.Backlight 175] Exception while processing button callback: AttributeError("'Backlight' object has no attribute 'has_xbacklight'")
2019-05-27 23:05:53,754 [CRITICAL][i3pystatus.backlight.Backlight 176] Traceback (most recent call last):
  File "/nix/store/v0km55axl2452hnsazbc1y8k0zdmcsbv-i3pystatus-unstable-2019-02-10/lib/python3.7/site-packages/i3pystatus/core/modules.py", line 162, in __button_callback_handler
    call_callback(getattr(self, cb), *args, **kwargs)
  File "/nix/store/v0km55axl2452hnsazbc1y8k0zdmcsbv-i3pystatus-unstable-2019-02-10/lib/python3.7/site-packages/i3pystatus/core/modules.py", line 133, in call_callback
    cb(*args, **kwargs)
  File "/nix/store/v0km55axl2452hnsazbc1y8k0zdmcsbv-i3pystatus-unstable-2019-02-10/lib/python3.7/site-packages/i3pystatus/backlight.py", line 93, in darker
    if self.has_xbacklight:
AttributeError: 'Backlight' object has no attribute 'has_xbacklight'
```
- It allows the user to set `format_no_backlight` to empty string and hide `backlight` module in case the system does not have a backlight (i.e.:, again, in a Desktop). If you set to an empty string nowadays it results the same output as the default one, resulting in something like `Brightness: -1` output.

PR #749, opened 2019-06-01
Add light support to backlight module

thiagokokada:
Light (https://github.com/haikarainen/light) is a program to control backlight using sysfs files and udev rules instead of through a X protocol like xbacklight, so it works in non-X environments. Also, it seems more reliable and works where xbacklight does not work (like when using xf86-modesetting instead of xf86-video-intel driver).

It is already supported in multiple distros (like Arch, Fedora and NixOS), so it seems to be a much better supported alternative than other programs that do something similar (like https://github.com/wavexx/acpilight/).

Substitutes PR https://github.com/enkore/i3pystatus/pull/747.


thiagokokada:
Tested:
- Notebook, with `adjust_method` set to `xbacklight`, `light` and default (`None`).
- Desktop, with `format_no_backlight` set to empty string.

thiagokokada:
@enkore

johnliu55tw:
Can anyone help review this? `xbacklight` only works with Intel (according to [ArchWiki](https://wiki.archlinux.org/index.php/Backlight#xbacklight)) so I can't utilize this plugin to adjust my AMD laptop. Thanks!

PR #750, opened 2019-06-07
pulseaudio: remove output when sink is gone

sileht:
When `sink` is set and you have USB card or Bluetooth headset.

When the sink is created, the output is set and the module appear.

But when the sink is deleted, the output is not deleted, and the module
stay.

This change detects when the sink is deleted and delete the output.

### DIFF
--- a/i3pystatus/pulseaudio/__init__.py
+++ b/i3pystatus/pulseaudio/__init__.py
@@ -147,7 +147,7 @@ class PulseAudio(Module, ColorRangeModule):
 
         self.request_update(context)
 
-    def sink_info_cb(self, context, sink_info_p, _, __):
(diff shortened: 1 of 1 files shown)


i3pystatus/pulseaudio/__init__.py
```diff
@@ -147,7 +147,7 @@ def update_cb(self, context, t, idx, userdata):
 
         self.request_update(context)
 
-    def sink_info_cb(self, context, sink_info_p, _, __):
+    def sink_info_cb(self, context, sink_info_p, eol, __):
```
thiagokokada: You can change this `__` to a single `_`.

sileht: done

chestm007:
closing and reopening to trigger a rebuild

PR #761, opened 2019-09-06
Bluetooth Module

gdunstone:
I made an extremely simple bluetooth module.

Its not what I would call complete, but its a small start that is functional enough to be somewhat useful.


enkore:
Thanks

PR #762, opened 2019-09-27
Add a new module to display charge status of a Tesla

drwahl:
This module will use the Tesla API to display the charge information for your Tesla vehicle.

### DIFF
--- a/i3pystatus/teslacharge.py
+++ b/i3pystatus/teslacharge.py
@@ -1,0 +1,124 @@
+from tesla_api import TeslaApiClient
+from i3pystatus import IntervalModule
+
+
+class TeslaCharge(IntervalModule):
+    """
+    Displays the current charge/range of your Tesla vehicle. There is a ton of
+    data that could be displayed, so read this module to see the full list of
+    datapoints that can be displayed.
+    Requires: tesla_api
+    """
+
+    settings = (
+        ("email", "Email address used to login to your Tesla account."),
+        ("password", "Password for accessing your Tesla account."),
+        ("units", "Miles (default) or km"),
+        ("interval", "Poll frequency. Polling too often can drain the battery"),
+        ("format_with_charge", "Display format while car is charging"),
+        ("format_without_charge", "Display format while car is not charging")
+    )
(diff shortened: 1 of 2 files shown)


terminalmage:
This breaks CI because Travis doesn't have the `tesla_api` module installed. It should probably be gated in some way and then an error raised if you try to _use_ the module without `tesla_api` installed. Or the import could be moved into an `init()` function.

@enkore What are your thoughts on this?

enkore:
See https://i3pystatus.readthedocs.io/en/latest/module.html#handling-dependencies

You'd just need to add "tesla_api" to MOCK_MODULES in docs/conf.py

i3pystatus/teslacharge.py
```diff
@@ -0,0 +1,123 @@
...
+    datapoints that can be displayed.
+    Requires: tesla_api
+    """
+
+    settings = (
+        ("email", "Email address used to login to your Tesla account."),
+        ("password", "Password for accessing your Tesla account."),
+        ("units", "Miles (default) or km"),
+        ("interval", "Poll frequency. Polling too often can drain the battery"),
+        "format"
```
enkore: You're not using this one

i3pystatus/teslacharge.py
```diff
@@ -0,0 +1,123 @@
...
+    units = "miles"
+    conversion_factor = 0.62137119
+    charge_complete = "#00FF00"  # green
+    charging = "#FFFF00"         # yellow
+    offline_color = "#FF0000"    # red
+    interval = 900
+    charging_icon = '⚡'
+    disconnect_icon = ''
+    format_with_charge = "{name}: {charge_state_icon}{charge_rate} {battery_level}%/{charge_limit_soc}% ({battery_range})"
+    format_without_charge = "{name}: {battery_level}%/{charge_limit_soc}% ({battery_range})"
```
enkore: Only these two -- add them to `settings` above and you should be good to go

That's the cause of

> Module <TeslaCharge> has non-required setting(s) <format> with no default!

enkore:
The future is _naow_.

PR #768, opened 2019-10-26
Add glyph bar type to mem_bar

facetoe:


facetoe:
Good point, I'll fix er up

PR #769, opened 2019-11-16
Allowing to set the duration format for the timewarrior module

tassos:
Currently, only one format is available for the timewarrior module, and it is not a very convenient one. By adding one row, the duration_format setting is changable from the configuration file.

I am not sure if I had to update the documentation, I suppose is it automatically generated from the code, right?


tassos:
Any comments on this? Would you be interested in accepting such a pull request?

terminalmage:
Yeah, this looks fine. Sorry no one saw this before.

PR #773, opened 2020-02-26
Fix the spotify module

grazzolini:
The now_playing module does not use player_name anymore so the spotify module is broken.
Change it to use player so it works correctly.

### DIFF
--- a/i3pystatus/spotify.py
+++ b/i3pystatus/spotify.py
@@ -5,4 +5,4 @@ class Spotify(NowPlaying):
     """
     Get Spotify info using dbus interface. Based on `now_playing`_ module.
     """
-    player_name = "spotify"
+    player = "spotify"


enkore:
Good find, thanks!

terminalmage:
Yeah this is interesting, it looks like `now_playing` never used `player_name`, so this has apparently not been working since 5e3434bd297f607a63641d7a29397133ac368b70, 3 years ago. Guess very few people actually use this module. :laughing:

grazzolini:
@terminalmage 
Actually I have been using the git version of i3pystatus for a while and I only noticed this module breaking last week.

Also, it's not that it wasn't working, but it would show an error, even with hide_no_player=True. If *any* player was present, it would show on the bar, which was weird. @enkore I would suggest to drop this module entirely in favor of now_playing.

PR #777, opened 2020-04-10
Update Github module to pass access token in request header

### LINKED ISSUE #772: Github module needs to be updated to reflect deprecation in API
See https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters

I should be able to get to this pretty soon.

terminalmage:
Also update the status colors to match the current Github Status webpage CSS.

Resolves #772.


terminalmage:
I'm seeing intermittent credential errors and need to check if it's just a bad access token

terminalmage:
It wasn't a bad access token, it was `requests` being a jerk and rewriting the `Authorization` header. I had to find an alternate means of passing the token.

terminalmage:
The issue was caused by a .netrc file that I didn't realize I had.

PR #781, opened 2020-05-07
Add wireguard module

Pluggi:
This PR adds a module for monitoring wireguard connections. It is basically a copy-paste of the openvpn module.


i3pystatus/wireguard.py
```diff
@@ -0,0 +1,81 @@
+from i3pystatus import IntervalModule
+from i3pystatus.core.command import run_through_shell
+
+__author__ = 'Pluggi'
+
+
+class OpenVPN(IntervalModule):
```
terminalmage, line 7: Probably want to change this class name.

Pluggi: Done! :D

i3pystatus/wireguard.py
```diff
@@ -0,0 +1,81 @@
...
+            color, status = self.color_up, self.status_up
+        else:
+            color, status = self.color_down, self.status_down
+
+        vpn_name = self.vpn_name
+        label = self.label
+
+        self.data = locals()
+        self.output = {
+            "full_text": self.format.format(**locals()),
```
terminalmage, line 79: If you're already assigning the locals to self.data, probably best to use `**self.data` here.

i3pystatus/wireguard.py
```diff
@@ -0,0 +1,81 @@
...
+
+    """
+
+    color_up = "#00ff00"
+    color_down = "#FF0000"
+    status_up = '▲'
+    status_down = '▼'
+    format = "{vpn_name} {status}"
+
+    status_command = "bash -c 'systemctl show wg-quick@%(vpn_name)s | grep ActiveState=active'"
```
xdbob, line 31:
```
/usr/bin/systemctl is-active wg-quick@%(vpn_name)s
```

?

Pluggi:
@terminalmage could you please review this again?

i3pystatus/wireguard.py
```diff
@@ -0,0 +1,81 @@
...
+            command = self.vpn_down_command
+        else:
+            command = self.vpn_up_command
+        run_through_shell(command % {'vpn_name': self.vpn_name})
+
+    def on_click(self, button, **kwargs):
+        self.toggle_connection()
+
+    def run(self):
+        command_result = run_through_shell(self.status_command % {'vpn_name': self.vpn_name})
```
terminalmage: This seems needlessly complicated. Also, C-style format strings are not very common in Python in 2021. It would arguably be more user-friendly for the end-user to create a format string like so:

`systemctl is-active wg-quick@{vpn_name}`

Then to process the format string you can use `self.status_command.format(vpn_name=self.vpn_name)`

This same pattern is repeated in a few places in this module with the different commands. I think they should all be changed to use the more user-friendly format strings.

terminalmage (changes requested):

i3pystatus/wireguard.py
Pluggi: I just pushed a fix for it

enkore:
This looks good to me 👍

PR #784, opened 2020-06-27
Fix "NowPlaying: KeyError: None"

AnyTimeTraveler:
This is a bad workaround, as it only catches the exception and doesn't address the problem at the place where it's thrown.
But I don't have time to study the code properly, as it's quite late in the evening right now.

If someone else wants to improve on this before I get to it, feel free to.


lukasvdberk:
can this be added please

PR #785, opened 2020-07-16
Add a new module to manage Yubikey devices

dth0:
This PR allows Yubikey's users to disable and re-enable their keys whenever they want in order to avoid triggering the OTP accidentally, It is also possible to create a python script to active the key and use it with a shortcut.

### DIFF
--- a/i3pystatus/yubikey.py
+++ b/i3pystatus/yubikey.py
@@ -1,0 +1,139 @@
+import re
+import os
+import time
+
+from i3pystatus import IntervalModule
+from i3pystatus.core.command import run_through_shell
+
+
+class Yubikey(IntervalModule):
+    """
+    This module allows you to lock and unlock your Yubikey in order to avoid
+    the OTP to be triggered accidentally.
+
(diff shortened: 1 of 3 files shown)


i3pystatus/yubikey.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self):
+        super().__init__()
+
+    @property
+    def _device_id(self):
+        command = run_through_shell(
+            "xinput list",
+            enable_shell=True
+        )
```
enkore, line 54: Generally I find not running things on a shell preferable, if not necessary by design. xinput should always be the same, regardless of shell preferences, so just leaving enable_shell off should work.

enkore:
Thanks, looks good to me.

i3pystatus/yubikey.py
dth0: tks for the feedback, I just push a new commit change these lines as you suggested.

enkore:
Neat, thanks for fixing that Travis problem as well :)

PR #788, opened 2020-07-30
Update weather.com backend to reflect upstream changes

terminalmage:
Weather.com embeds the weather data as a JSON in a `<script>` tag, so this just updates to reflect recent changes to how that information is embedded.

### DIFF
--- a/i3pystatus/weather/weathercom.py
+++ b/i3pystatus/weather/weathercom.py
@@ -70,9 +70,12 @@ class WeathercomHTMLParser(HTMLParser):
                 weather_data = None
                 self.logger.debug('Located window.__data')
                 # Strip the "window.__data=" from the beginning and load json
-                json_data = self.load_json(
-                    content[begin:].split('=', 1)[1].lstrip()
-                )
+                raw_json = content[begin:].split('=', 1)[1].lstrip()
+                if re.match(r'^JSON\.parse\("', raw_json):
+                    raw_json = re.sub(r'^JSON\.parse\("', '', raw_json)
+                    raw_json = re.sub(r'"\);?$', '', raw_json)
+                    raw_json = raw_json.replace(r'\"', '"').replace(r'\\', '\\')
+                json_data = self.load_json(raw_json)
                 if json_data is not None:
                     try:
                         weather_data = json_data['dal']


howlett:
Does this work for international data?  I'm in canada and this commit stops my weather from being loaded.  I updated because the status was always "None", but now I don't get anything.

howlett:
How can I help with this? I've had no success with enabling debug in this module as described in the pid file.

terminalmage:
@howlett I do this:

```python
# Adjust this logfile path as needed
status = Status(logfile='/home/erik/var/i3pystatus.log')

status.register(
    'weather',
    hints={'markup': 'pango'},
    format='{condition} {current_temp}{temp_unit}[ {icon}][ Hi: {high_temp}][ Lo: {low_temp}][ {update_error}]',
    colorize=True,
    refresh_icon='<span color="' + COLOR_WORKING + '">⟳</span>',
    log_level=logging.DEBUG,
    backend=weathercom.Weathercom(
        units='imperial',
        location_code='whatever_you_use',
        log_level=logging.DEBUG,
    )
)
```

It also helps to run i3pystatus from a virtualenv, because it lets you install it using `pip --editable /path/to/git/clone/of/i3pystatus` to make it easier to run i3pystatus while you are developing.

Sorry for any inconvenience, and I hope you're able to find the issue. Let me know if I can be of any further assistance.

terminalmage:
@howlett The `None` issue should be fixed in https://github.com/enkore/i3pystatus/pull/789. This also makes fixes to reflect yet another API change.

howlett:
Thanks.  I have tried to upgrad with "sudo pip3 install git+
https://github.com/enkore/i3pystatus.git --upgrade", but it seems I still
get 3.35.  Is this correct?  I still get nothin but "() !" in the status
bar after running the upgrade.

{~} $ python --version
Python 2.7.18
{~} $ python3 --version
Python 3.8.5
{~} $

Regards,
Liam

terminalmage:
3.35 is the last numbered release. Since then we just recommend installing from git (I'm going to be working on improving docs this week).

I personally don't use pip to install i3pystatus globally, I use a virtualenv.

```bash
virtualenv /path/to/my/virtualenvs/i3pystatus
pip install --editable /path/to/git/clone/of/i3pystatus
```

Then invoke your status script via the virtualenv's copy of the python binary, by setting the following in `$HOME/.config/i3/config`

```
status_command /path/to/my/virtualenvs/i3pystatus/bin/python /path/to/my/status_script.py
```

If you do this, updating is just as easy as pulling the git repo and then reloading i3.

PR #791, opened 2020-09-30
Add new status "maintenance" for github module

terminalmage:
See https://www.githubstatus.com/incidents/ry5psbndh11j

Also, don't configure defaults for the status text, and just fall back to GitHub if there is no status text defined for the current status code.


enkore:
Nice find

Issue #786, opened 2020-07-17
'window_title' module crashes error


prietus:
i3pystatus version: i3pystatus-git 3.35.r346.gdad5eb0-1
os: arch linux 
wm: i3-wm 4.18.1-1

status.register(
    'window_title',
    format="{class_name} {title}"
)
error:
ERROR:Could not Parse JSON (lexical error: invalid char in jsin text.))

terminalmage:
I'm unable to reproduce, is this still happening?

prietus:
it is. if I delete this part of the code (diff attached), the issue gets fixed. Not saying that's a solution but it can help to clarify the issue.
[issue.diff.txt](https://github.com/enkore/i3pystatus/files/5141320/issue.diff.txt)

terminalmage:
OK, well that code is pretty important to the module. One way you might be able to troubleshoot is to run the actual code that the module is using and see where it breaks down. For example, copy the following into a file `foo.py` and then run `python foo.py`:

```python
import i3ipc
conn = i3ipc.Connection()
tree = conn.get_tree()
w = tree.find_focused()
p = w.parent()
print('w.border = {}'.format(w.border))
print('w.type = {}'.format(w.type))
print('p.layout = {}'.format(p.layout))
print('len(p.nodes) = {}'.format(len(p.nodes)))
```

prietus:
I get:
Traceback (most recent call last):
  File "foo.py", line 7, in <module>
    p = w.parent()
TypeError: 'Con' object is not callable

terminalmage:
OK, this seems like an upstream issue, can you report it to https://github.com/altdesktop/i3ipc-python?

PR #799, opened 2020-11-24
window_title: Use empty_title if title is None

ghost:
Uses the default empty title for windows that do not have a title, preventing a TypeError in the max_width check.

The "View Page Info..." dialog in Firefox is an example of a window that starts off with a null title until it is fully loaded, at least in sway.


ghost:
Likely fixes #786

PR #800, opened 2020-11-25
Fix API error

tyjak:
- Replace the old API bitcoinaverage.com by bitaps.com
- Add bistamp as default exchange
- remove and replace some fields to adapt the new API
- Replace Bitcoin symbol by \uF15A


tyjak:
not the new current default branch...

PR #808, opened 2021-02-25
Adding new hassio modulue

drwahl:
Initial commit for a new HomeAssistant.io module to report the state of
entities from hassio.

Signed-off-by: David Wahlstrom <david.wahlstrom@gmail.com>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -68,8 +68,7 @@ MOCK_MODULES = [
     "requests.adapters",
     "exchangelib",
     "soco",
-    "tesla_api",
-    "json"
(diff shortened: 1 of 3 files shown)


enkore:
Looks good, but I think you need to add a dependency mock, see https://i3pystatus.readthedocs.io/en/latest/module.html#handling-dependencies

drwahl:
I'm having a tough time getting this CI job to pass. It's complaining about the "requests" module, but I added a bit to the docstring about needing it, and it's already included in MOCK_MODULES. What am I missing?

PR #818, opened 2021-06-03
Replace pacmd calls with pactl to support Pipewire

### LINKED ISSUE #810: Support for pipewire
I have switched from pulseaudio to pipewire and the pulseaudio module doesn't seem to work even as a pulse client/app with pipewire. I suspect that the wrapping done by pulse.py breaks down when using pipewire-pulse.

EgZvor:
`pacmd` does not work with Pipewire, use `pactl` where possible instead

Fixes #810

### DIFF
--- a/i3pystatus/pulseaudio/__init__.py
+++ b/i3pystatus/pulseaudio/__init__.py
@@ -197,13 +197,15 @@ class PulseAudio(Module, ColorRangeModule):
             else:
                 raise Exception("bar_type must be 'vertical' or 'horizontal'")
 
-            selected = ""
-            dump = subprocess.check_output("pacmd dump".split(), universal_newlines=True)
-            for line in dump.split("\n"):
-                if line.startswith("set-default-sink"):
-                    default_sink = line.split()[1]
-                    if default_sink == self.current_sink:
-                        selected = self.format_selected
+            # TODO: Looks like we can't access daemon config using pactl.
+            # Remove this part if that's true or find an appropriate pactl alternative.
+            # selected = ""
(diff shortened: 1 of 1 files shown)


i3pystatus/pulseaudio/__init__.py
```diff
@@ -197,22 +197,23 @@ def sink_info_cb(self, context, sink_info_p, eol, _):
...
-                        selected = self.format_selected
+            # TODO: Looks like we can't access daemon config using pactl.
+            # Remove this part if that's true or find an appropriate pactl alternative.
+            # selected = ""
+            # dump = subprocess.check_output("pacmd dump".split(), universal_newlines=True)
+            # for line in dump.split("\n"):
+            #     if line.startswith("set-default-sink"):
+            #         default_sink = line.split()[1]
+            #         if default_sink == self.current_sink:
+            #             selected = self.format_selected
```
hasB4K, line 208: I'm pretty sure that this issue can be easily be resolved with the following:
```python
            selected = ""
            default_sink = subprocess.check_output("pactl get-default-sink".split()).strip()
            if default_sink == self.current_sink:
                selected = self.format_selected
```

hasB4K: I have tested with a full pulseaudio and a full pipewire setup, and this patch works.

`pactl get-default-sink` and `pacmd dump | grep set-default-sink | awk '{print $2}'` (the equivalent bash command of the current code) gives the same outputs on a full pulseaudio setup.

hasB4K:
Thank you @EgZvor for this patch! I think this should be merged (with the small patch that I provided). It works perfectly for me.

PR #819, opened 2021-06-18
mpd: add option "time_format" to module

fkusei:
This allows customizing the time display in MPD module.

For example, i'm usually playing long (2+ hour) mixes of some DJs, so only having %M:%S is not sufficient for me.

Default behaviour is unchanged.

I know one could also use https://i3pystatus.readthedocs.io/en/latest/configuration.html#timewrapper, but I think *one* setting for both is better.


enkore:
Thanks!

PR #825, opened 2021-09-20
GSSAPI support for IMAP

tukss:
This adds GSSAPI support for the IMAP mail plugin. It is useful in situations where the mail server supports passwordless authentication via Kerberos, especially when there is no password in the first place.

In it's current state the password field is still required, but can be set to any, including empty, value, when GSSAPI is active.

The code was adapted from a code snippet by Yves Fischer.
https://gist.github.com/yvesf/7cc8300176df8322c913
@yvesf are you OK with this code being published under the MIT license as part of i3pystatus?

### DIFF
--- a/i3pystatus/mail/imap.py
+++ b/i3pystatus/mail/imap.py
@@ -9,12 +9,48 @@ except ImportError:
 import contextlib
 import time
 import socket
+import base64
 from threading import Thread
 
 from i3pystatus.mail import Backend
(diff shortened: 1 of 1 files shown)


yvesf:
@tukss no problem, I don't mind

tukss:
I just rebased on `current`. Please let me know if you want any other changes to get this merged.

PR #836, opened 2022-02-04
Replace pacmd calls with pactl to support Pipewire [final version]

### LINKED ISSUE #810: Support for pipewire

hasB4K:
Hello,

I continued the work from @EgZvor (thank you!) in #818 and patched the remaining issue with default-sink in pactl.
I used this version for the last two weeks so far. I think this PR is directly megeable @enkore :wink: 

Fixes: #810 

Best regards,
hasB4K

### DIFF
--- a/i3pystatus/pulseaudio/__init__.py
+++ b/i3pystatus/pulseaudio/__init__.py
@@ -1,4 +1,3 @@
-import os
 import re
 import subprocess
 
@@ -46,7 +45,7 @@ class PulseAudio(Module, ColorRangeModule):
         ("format_muted", "optional format string to use when muted"),
         ("format_selected", "string used to mark this sink if selected"),
         "muted", "unmuted",
-        "color_muted", "color_unmuted",
+        "color_error", "color_muted", "color_unmuted",
         ("step", "percentage to increment volume on scroll"),
         ("sink", "sink name to use, None means pulseaudio default"),
         ("move_sink_inputs", "Move all sink inputs when we change the default sink"),
@@ -64,6 +63,7 @@ class PulseAudio(Module, ColorRangeModule):
     format_selected = " 🗸"
     currently_muted = False
     has_amixer = False
+    color_error = "#FF0000"
     color_muted = "#FF0000"
     color_unmuted = "#FFFFFF"
     vertical_bar_glyphs = None
@@ -120,8 +120,7 @@ class PulseAudio(Module, ColorRangeModule):
         if self.sink is not None:
             return self.sink
 
-        self.sinks = subprocess.check_output(['pactl', 'list', 'short', 'sinks'],
-                                             universal_newlines=True).splitlines()
+        self.sinks = self._call_pactl(["list", "short", "sinks"]).splitlines()
         bestsink = None
         state = 'DEFAULT'
         for sink in self.sinks:
@@ -191,19 +190,17 @@ class PulseAudio(Module, ColorRangeModule):
                 output_format = self.format
 
             if self.bar_type == 'vertical':
-                volume_bar = make_vertical_bar(volume_percent, self.vertical_bar_width, glyphs=self.vertical_bar_glyphs)
+                volume_bar = make_vertical_bar(volume_percent, self.vertical_bar_width,
+                                               glyphs=self.vertical_bar_glyphs)
             elif self.bar_type == 'horizontal':
                 volume_bar = make_bar(volume_percent)
             else:
                 raise Exception("bar_type must be 'vertical' or 'horizontal'")
 
             selected = ""
-            dump = subprocess.check_output("pacmd dump".split(), universal_newlines=True)
-            for line in dump.split("\n"):
-                if line.startswith("set-default-sink"):
-                    default_sink = line.split()[1]
-                    if default_sink == self.current_sink:
-                        selected = self.format_selected
+            default_sink = self._call_pactl(["get-default-sink"]).strip()
+            if default_sink == self.current_sink:
+                selected = self.format_selected
 
             self.output = {
                 "color": color,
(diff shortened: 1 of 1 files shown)


hasB4K:
@terminalmage do you have the rights to merge this PR? :)

i3pystatus/pulseaudio/__init__.py
```diff
@@ -198,12 +198,9 @@ def sink_info_cb(self, context, sink_info_p, eol, _):
                 raise Exception("bar_type must be 'vertical' or 'horizontal'")
 
             selected = ""
-            dump = subprocess.check_output("pacmd dump".split(), universal_newlines=True)
-            for line in dump.split("\n"):
-                if line.startswith("set-default-sink"):
-                    default_sink = line.split()[1]
-                    if default_sink == self.current_sink:
-                        selected = self.format_selected
+            default_sink = subprocess.check_output("pactl get-default-sink".split()).strip()
```
terminalmage: If pulseaudio is for some reason not running, and `pactl` is run, will it return a nonzero exit code? Because if so, `subprocess.check_output()` will raise an exception that you are not catching here.

Queueue0: The same issue exists with pacmd and the original code doesn't catch the exception either. I ran with both this version and the version with pacmd and got the same output when pulse wasn't running. It appears to just run as if the pulseaudio module isn't registered.

hasB4K: Hi,

Apologies for the delay in responding to this PR. I received an email notification a day ago regarding a comment from a GitHub user who pointed out that the current code is already causing an error if `pacmd` is not present. However, it appears that the user has since deleted his comment.

In my view, hiding this error might not be beneficial for end users, as it could indicate a configuration issue with `pactl`. I propose three options:
1. Keep the code as is, assuming users will investigate their system configuration.
2. Display the error differently.
3. Hide the error entirely, though I believe this is not ideal.

I'd like this PR to be merged. @terminalmage, what's your preference among these options? Or maybe you would like to something else entirely?

Kind regards,

terminalmage: My preference is to write the error to the logger, and display something indicating that there is an error in the module text. I am a big fan of using a logfile in the i3pystatus script, since the string version of the traceback (which is displayed when there is an uncaught exception) is sometimes unhelpful. For example, it's good at identifying that there is a missing dependency that couldn't be imported, but you also occasionally get an error that says something like `TypeError: expected int, got NoneType`.

I think the [configuration docs](https://i3pystatus.readthedocs.io/en/latest/configuration.html) should probably be updated to add a `logfile` argument when instantiation your `Status` instance, but that's another discussion entirely (maybe I'll open a PR for that :smile:).

Each `IntervalModule` subclass instance already has a logger assigned to an attribute, so you could use something like:

```python
self.logger.error("some error")
```

If you are logging from within an `except` block, then you can use `self.logger.exception("some error")`, which is equivalent to `self.logger.error("some message", exc_info=True)`. This adds the original traceback to the log file, and logs at the `ERROR` loglevel.

i3pystatus/pulseaudio/__init__.py
```diff
@@ -223,28 +220,33 @@ def sink_info_cb(self, context, sink_info_p, eol, _):
...
-            sink_inputs = subprocess.check_output("pacmd list-sink-inputs".split(),
-                                                  universal_newlines=True)
+            sink_inputs = self._call_pactl(["list-sink-inputs"])
             for input_index in re.findall(r'index:\s+(\d+)', sink_inputs):
-                command = "pacmd move-sink-input {} {}".format(input_index, next_sink)
-
-                # Not all applications can be moved and pulseaudio, and when
-                # this fail pacmd print error messaging
-                with open(os.devnull, 'w') as devnull:
-                    subprocess.call(command.split(), stdout=devnull)
```
hasB4K, line 240: This part of the code came from https://github.com/enkore/i3pystatus/pull/552. Since this PR uses the method `_call_pactl` instead that internally uses `subprocess.check_output`, the messages on stdout that could come from `pactl` will always be discarded.

i3pystatus/pulseaudio/__init__.py
```diff
@@ -223,28 +220,33 @@ def sink_info_cb(self, context, sink_info_p, eol, _):
...
+    def _call_pactl(self, pactl_arguments):
+        try:
+            output = subprocess.check_output(
+                ["pactl"] + [str(arg) for arg in pactl_arguments],
+                universal_newlines=True)
+            return output
+        except Exception as exc:
+            self.logger.exception("Error while executing pactl")
+            self.output = {"color": self.color_error, "full_text": "Error while executing pactl"}
+            self.send_output()
```
hasB4K: Every call of the functions `subprocess.check_output` or `subprocess.call` of this module were replaced in this PR by `self._call_pactl`.
This method call the `pactl` command with its arguments using `subprocess.check_output()` internally. The outputs are either treated by the callee or simply ignored.

If an error occurs, we update `self.output` with a minimal message and we notify the change using `self.send_output()`. The full stacktrace is available in the file logs (I checked) since `self.logger.exception` is now used.

hasB4K:
@terminalmage let me know if the changes that I made are corresponding to what you had in mind :slightly_smiling_face: 

FYI, when `pactl` is not present, this is what is now displayed in i3bar:
Screenshot of the error

i3pystatus/pulseaudio/__init__.py
```diff
@@ -223,28 +220,33 @@ def sink_info_cb(self, context, sink_info_p, eol, _):
...
+                self._call_pactl(pactl_args)
+        self._call_pactl(["set-default-sink", next_sink])
+
+    def _call_pactl(self, pactl_arguments):
+        try:
+            output = subprocess.check_output(
+                ["pactl"] + [str(arg) for arg in pactl_arguments],
+                universal_newlines=True)
+            return output
+        except Exception as exc:
```
terminalmage: Sorry for the nitpick on this, but since you're not referencing `exc` in the `except` block, it's not necessary to assign it to a variable. Doing so is not required for the logger to gain access to the exception's context.

terminalmage (changes requested):
Just one remaining small nitpick.

i3pystatus/pulseaudio/__init__.py
hasB4K: Thanks for the review :slightly_smiling_face:, it should be fixed now! I actually forgot to remove this variable from a previous debug...

hasB4K:
@terminalmage Thanks for the review! Let me know if I need to fix something else :slightly_smiling_face:

terminalmage (approved):

terminalmage:
Thanks! Merged it just now.

PR #838, opened 2022-03-12
Add docker module to report on docker assets

drwahl:
Signed-off-by: David Wahlstrom <david.wahlstrom@dreamhost.com>

### DIFF
--- a/i3pystatus/docker.py
+++ b/i3pystatus/docker.py
@@ -1,0 +1,50 @@
+import docker
+from i3pystatus import IntervalModule
+
+
+class Docker(IntervalModule):
+    """
+    Display information about docker containers
+    Requires: docker
+    """
+
+    settings = (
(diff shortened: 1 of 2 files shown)


i3pystatus/docker.py
```diff
@@ -0,0 +1,50 @@
+import docker
+from i3pystatus import IntervalModule
+
+
+class Docker(IntervalModule):
+    """
+    Display information about docker containers
+    Requires: docker
+    """
```
terminalmage, line 9: Can you add some documentation here about the available formatters?

terminalmage (changes requested):

chestm007:
this errors out when no containers are running as containers['running'] isnt populated. ive hacked around it be setting that to 0 for now before the output is calculated, but it feels dirty

chestm007:
merging, ill push a patch to clear up the issue shortly after

PR #841, opened 2022-05-16
Fix default MPD time format

adralioh:
Change the default value of the `time_format` setting in the mpd module so that it matches the default value used by `TimeWrapper`.

This fixes a bug where the seconds were incorrectly displaying as `MS` by default due to a typo.

Before: `02:MS`
After: `2:06`

Reference: #819


terminalmage:
Good catch.

PR #842, opened 2022-07-03
Add WIFIonICE module

Kunsi:
This module adds support for displaying informations about the current trip on Deutsche Bahn high speed trains in i3pystatus. On click, travelynx (a popular trip-tracking tool) will open to allow for easy check in into the train.

TODOs:
- [x] Add documentation
- [x] Make output configurable to be able to show different parts of the information

### DIFF
--- a/i3pystatus/wifionice.py
+++ b/i3pystatus/wifionice.py
@@ -1,0 +1,176 @@
+from datetime import datetime, timedelta
+from urllib.request import urlopen
+from subprocess import check_output
+from json import loads
+
+from i3pystatus import IntervalModule
+from i3pystatus.core.util import internet, require
+
+
+class WifiOnIceAPI(IntervalModule):
+    """
+    Displays information about your current trip on Deutsche Bahn trains.
+
+    Requires the PyPI package `basiciw` if you want to use automatic
+    detection. See below on how to disable automatic detection based on
+    wifi network names.
+
+    .. rubric:: Available formatters
+
+    For `format_delay` option:
+
+    * `{delay}` - delay of train in minutes
+
+    For `format_network_info` option:
+
+    * `{currently}` - current state of network quality
+    * `{duration}` - how long until the next network quality change
+    * `{expected}` - next state of network quality
+
+    For `format_next_stop` option:
+
+    * `{station}` - Station name
+    * `{platform}` - Platform number or name
+    * `{arrival_time}` - Arrival time of train at the station (actual, if available, otherwise scheduled)
+    * `{arrival_in}` - Time until arrival (in form "1h 12m" or "53m")
+    * `{delay}` - Output of `format_delay` (or empty if no delay)`
+
+    For `format_output` formatter:
+
+    * `{speed}` - Train speed in km/h
+    * `{next_stop}` - Output of `format_next_stop` (or `final_destination` option)
+    * `{network_info}` - Output of `format_network_info` (or empty if neither current nor next info is known)
+    """
+
+    final_destination = 'Endstation, bitte aussteigen!'
+    format_delay = ' | <span color="#FF0000">{delay_minutes}</span>'
+    format_network_info = ' <span color="#999999">(Net: {currently} > [{duration}] {expected})</span>'
+    format_next_stop = '{station} <span color="#999999">[{platform}]</span> {arrival_time} ({arrival_in}{delay})'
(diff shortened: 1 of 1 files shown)


Kunsi:
This should now be ready to merge, will do a final test on wednesday when i'm on a train again

Kunsi:
Is there anything else needed from my side to get this merged?

terminalmage:
Sorry for the long delay... did you have any other work you wanted to do here?

Kunsi:
No, this should be ready to merge.

i3pystatus/wifionice.py
```diff
@@ -0,0 +1,219 @@
...
+            minutes = int(seconds / 60)
+            seconds -= minutes * 60
+            components.append('{}m'.format(minutes))
+        if not components:
+            components.append('now')
+        return " ".join(components)
+
+    def _check_wifi(self):
+        if self.wifi_adapters is None:
+            self.logger.debug('self.wifi_adapters is None')
```
terminalmage: This log message could be improved. Maybe something like "automatic on-train detection is disabled"

i3pystatus/wifionice.py
```diff
@@ -0,0 +1,219 @@
...
+                    self.trip_info = loads(trip_info_req.read())['trip']
+
+                    ice_status_req = urlopen('https://iceportal.de/api1/rs/status')
+                    self.ice_status = loads(ice_status_req.read())
+                except Exception:
+                    self.trip_info = {}
+                    self.ice_status = {}
+
+                self.logger.debug(f'trip_info: {repr(self.trip_info)}')
+                self.logger.debug(f'ice_status: {repr(self.ice_status)}')
```
terminalmage: Python string formatting has built-in support for invoking `repr`, using `!r`. So, these could both be done using:

```python
self.logger.debug(f'trip_info: {self.trip_info!r}')
self.logger.debug(f'ice_status: {self.ice_status!r}')
```

i3pystatus/wifionice.py
```diff
@@ -0,0 +1,219 @@
...
+                with self.condition:
+                    self.condition.wait(self.on_train_interval)
+            else:
+                self.logger.info('Not on a train :(')
+
+                self.trip_info = {}
+                self.ice_status = {}
+
+                self.logger.debug(f'trip_info: {repr(self.trip_info)}')
+                self.logger.debug(f'ice_status: {repr(self.ice_status)}')
```
terminalmage: Same as above, regarding python string formatting.

i3pystatus/wifionice.py
```diff
@@ -0,0 +1,219 @@
...
+        if not components:
+            components.append('now')
+        return " ".join(components)
+
+    def _check_wifi(self):
+        if self.wifi_adapters is None:
+            self.logger.debug('self.wifi_adapters is None')
+            return True
+
+        from basiciw import iwinfo
```
terminalmage: With the default configuration, this import will cause a traceback unless basiciw is installed.

My personal preference is not to make the default configuration result in a traceback, and instead communicate a misconfiguration (or missing dependencies) via logging. However, a number of other i3pystatus modules result in a traceback if there are missing dependencies, and the default behavior of the `Status` class doesn't use a log file.

@enkore Do you have a specific preference for handling missing dependencies?

i3pystatus/wifionice.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def _format_time(self, seconds):
+        if seconds is None:
+            return "?"
+        seconds = int(seconds)
+        components = []
+        if seconds >= 3600:
+            hours = int(seconds / 3600)
+            seconds -= hours * 3600
+            components.append('{}h'.format(hours))
```
terminalmage: This can be an f-string

i3pystatus/wifionice.py
```diff
@@ -0,0 +1,219 @@
...
+        if seconds >= 60:
+            minutes = int(seconds / 60)
+            seconds -= minutes * 60
+            components.append('{}m'.format(minutes))
```
terminalmage: This can also be an f-string

terminalmage (changes requested):

i3pystatus/wifionice.py
Kunsi: I have modified this to print a warning to the logs if basiciw is not installed.

Kunsi:
Sorry it took me so long, but this should be good to go now :)

PR #844, opened 2022-12-25
Migrate to inspect.getfullargspec()

stefanor:
getargspec() was deprecated in 3.0 and removed in 3.11.


terminalmage:
This looks good, and appears to be the only instance of `instance.getargspec()` in the codebase.

PR #855, opened 2024-01-21
remove dependency on old mock

a-detiste:


bparmentier:
Is something blocking this PR? It fixes the build error that (at least) Arch Linux users encounter.

terminalmage (approved):

PR #856, opened 2024-01-29
Fix get_api_key() on wunderground.py

elig0n:
The first API key match in the main webpage stopped working for my weather requests.
I used the second one instead and it appear to be working.

I counted 10 matches of `api_key=([0-9a-f]+)` altogether in the source, the 1st and 4th being the same and which appear to be a developers' API key. The others are the same key which seems to work.
In my patch I used the second match using finditer() and next()

### DIFF
--- a/i3pystatus/weather/wunderground.py
+++ b/i3pystatus/weather/wunderground.py
@@ -110,8 +110,10 @@ class Wunderground(WeatherBackend):
             self.logger.exception(f'Failed to load {url}')
         else:
             try:
-                return re.search(r'apiKey=([0-9a-f]+)', page_source).group(1)
-            except AttributeError:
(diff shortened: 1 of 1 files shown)


terminalmage:
Hey! Sorry for the delay, I've been away for a few days. I'll check this out soon.

Thanks very much for your time and effort.

terminalmage:
Hi! I took a look and ended up working up my own solution based on yours: https://github.com/enkore/i3pystatus/pull/858

This solution tries out each match returned by `re.finditer()` by making requests using that API key until it gets a proper response. The thinking behind this is that it will future-proof the code better against fluctuations in the appearance of `apiKey=.....` in the source.

I'm going to close this, but I wanted to re-iterate my thanks to you for reporting and correctly diagnosing the problem.

Cheers!

PR #860, opened 2024-03-14
Galaxy buds plugin

grazzolini:
Introduces a plugin to display information about Galaxy Buds devices using
    the earbuds tool from https://github.com/JojiiOfficial/LiveBudsCli

* Displays per bud battery level
* Displays case battery level
* Displays placement status
* Displays and sets both Ambient sound (AMB) and Active noise control (ANC)
* Has commands to connect/disconnect
* Dynamic coloring based on battery level
* Equalizer setting
* Touchpad lock

### DIFF
--- a/i3pystatus/buds.py
+++ b/i3pystatus/buds.py
@@ -1,0 +1,237 @@
+from enum import Enum, IntEnum
+from json import JSONDecodeError, loads
+from i3pystatus import IntervalModule
+from i3pystatus.core.command import run_through_shell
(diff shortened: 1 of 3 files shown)


grazzolini:
This is a plugin I have written for my own personal use and have been using for a while. I recently wrote tests for it and added a bunch of docstrings, so here is my PR. I'm not sure if I'm missing anything doc wise, but I added a ton of tests for it.

PR #865, opened 2024-04-22
Update the code for imp's removal with python 3.12

grazzolini:
* Changed the `__init__.py` file to use importlib instead of imp.


Panagiss:
We need this merge

grazzolini:
I've added this patch to the [i3pystatus](https://aur.archlinux.org/packages/i3pystatus-git) AUR package.

PR #868, opened 2024-05-17
Fix ping module with LANG fixed to C

tyjak:
Fix a pb when env language is different from LANG C


chestm007:
This PR looks like it has accidentally pulled in additional changes - could you clean it up please and ill take another look!

PR #869, opened 2024-06-18
Make sure buds tests can be ran from any path

grazzolini:
When running python setup.py pytest, the buds tests would fail due to the path. This fixes the relative path to come from the actual tests file.


lucasew (approved):

lucasew:
cc @terminalmage

PR #872, opened 2024-10-07
Add notifications to buds plugin and more

grazzolini:
* Added notifications when battery drift occurs and when one of the buds is on the case, when they reach same battery level.
* The notifications can be disabled by a configuration option.
* The symbols used to indicate the buds placement can be changed. We use regular characters, but printable unicode characters can be used too.
* Some code was refactored into functions to cleanup run()
* The tests were changed for the symbol change. Also a few tests for the notifications were added. Also, a few tests were extended to make sure the code changing the symbols work.


chestm007:
I have no way to test this, i assume you have :)
Happy to merge if it works, code looks pretty good to me

PR #873, opened 2024-10-20
Fix and refactor bitcoin module

tyjak:
* change API endpoint to blockchain.info
* remove wallet monitor functionality


chestm007:
Might need a little more info in this one, I'm not familiar with bitcoin anything at all, so please forgive any ignorance.
Why the change in exchange specified, and why remove the wallet monitor?

Issue #877, opened 2025-04-16
Battery module: DuplicateOptionError


4goettma:
After a recent system update on Arch Linux, the battery module outputs
`{"full_text": "BatteryChecker: DuplicateOptionError: While reading from '<string>' [line 17]: option 'TYPE' in section 'id10t' already exists", "color": "#FF0000", "name": "i3pystatus.battery.BatteryChecker", "instance": "136733527446192", "markup": "none"}`

The string "id10t" originates from https://github.com/enkore/i3pystatus/blob/current/i3pystatus/battery.py#L27
ChatGPT suggested to modify line 21 to `super().__init__(default_section="id10t", strict=False)` (adding `strict=False`) but I'm not sure if this is a reasonable fix regarding side effects (it works for me right now at least).

I'm using i3pystatus-git 3.35.r407.gf3c539a-5 [installed from the AUR](https://aur.archlinux.org/packages/i3pystatus-git).

I didn't managed yet to find the cause so here is the [list of recently upgraded packages](https://gist.githubusercontent.com/4goettma/c88d2dfe9b30bbf2da3d9b3024525e64/raw/5aea449278b26cd4f676d148c786da615fcb4c96/gistfile1.txt).

enkore:
Can you provide the uevent file? If you don't override this, it should be `/sys/class/power_supply/BAT*/uevent`. I don't think any relevant semantics have changed in ConfigParser for a long time, so it's probably the kernel outputting the TYPE line twice for some reason. strict=False is a valid workaround for this, but the responsible kernel module ought to be fixed anyway.

Chocobozzz:
Confirmed from `/sys/class/power_supply/BAT0/uevent`:

```
DEVTYPE=power_supply
POWER_SUPPLY_NAME=BAT0
POWER_SUPPLY_TYPE=Battery
POWER_SUPPLY_STATUS=Not charging
...
POWER_SUPPLY_CAPACITY_LEVEL=Normal
POWER_SUPPLY_TYPE=Battery
...
```

4goettma:
ThinkPad X220 with a single physical battery slot:

`/sys/class/power_supply/BAT0/uevent`:
```
DEVTYPE=power_supply
POWER_SUPPLY_NAME=BAT0
POWER_SUPPLY_TYPE=Battery
POWER_SUPPLY_STATUS=Discharging
POWER_SUPPLY_PRESENT=1
POWER_SUPPLY_TECHNOLOGY=Li-ion
[...]

PR #879, opened 2025-06-16
required until upstream fixes the duplicate battery type entry addresses #877

chestm007:
Signed-off-by: Max Chesterfield <max.chesterfield@zepben.com>


chestm007:
No ones chimed in, merging.

